PR #2, opened 2026-02-19
Simplification: OTLP Sidecar Observability

cl4wb0t:
## Summary

- **Simplify `lit-observability`**: remove legacy UDS/Proxy transport, `LitConfig` dependency, and `config` module. Standardize on OTLP/gRPC export with feature-gated (`otlp`/`production`) compilation.
- **Add OTLP Collector sidecar**: new `otel-collector/` with config, entrypoint, and `Dockerfile.otel-collector`. Receives OTLP/gRPC on `:4317`, exports to GCP Cloud Logging/Trace/Monitoring. Non-blocking dependency in `docker-compose.phala.yml` — services degrade to stdout if collector is unavailable.
- **Add `ObservabilityFairing`** (Rocket HTTP middleware): generates a UUID v4 `X-Request-Id` on every request (user-provided values are ignored), propagates optional `X-Correlation-Id` if the user sent one, attaches both headers to all responses (success, error, 404). Includes 8 tests covering all requirement edge cases.
- **Wire up provider init in `lit-api-server` and `lit-actions`**: read `RUST_LOG` and `LIT_TELEMETRY_ENDPOINT` from env vars, init tracing subscriber with optional OTLP layer, graceful provider shutdown on exit.
- **Context propagation**: `set_request_context` / `get_request_context` via span extensions + task-local storage (DashMap), plus `ContextAwareOtelLogLayer` that injects request/correlation IDs into OTel log records.
- **Observability requirements doc** (`architectureDocs/deployment/planning/observability/requirements.md`).

## Requirements coverage

Per `requirements.md`:

| Requirement | Implementation |
|---|---|
| Stdout/stderr logging if OTLP unavailable | `tracing-subscriber` always-on layer; OTLP feature-gated |
| OTLP export with required collector path | `LIT_TELEMETRY_ENDPOINT` env var, process exits if missing when `otlp` feature enabled |
| End-to-end request tracing with random span ID | `ObservabilityFairing` creates UUID v4 per request |
| `X-Request-Id` on all responses | Fairing `on_response` attaches to success, error, and 404 |
| User-provided `X-Request-Id` ignored | Fairing always generates fresh UUID (tested) |
| `X-Correlation-Id` propagated if user provides it | Fairing reads, stores, returns on response (tested) |
| `X-Correlation-Id` absent if user didn't send it | Fairing omits header from response (tested) |

## Test plan

- [ ] `cargo fmt --all -- --check` passes for all three crates
- [ ] `cargo clippy` passes for all three crates
- [ ] `cargo build` succeeds for `lit-api-server`, `lit-actions`, `lit-core`
- [ ] 8 fairing unit tests pass (`cargo test` in `lit-api-server`)
- [ ] Phala simulator validation passes
- [ ] Verify `X-Request-Id` and `X-Correlation-Id` headers in manual HTTP request/response


Garandor:
* **#2** <a href="https://app.graphite.com/github/pr/LIT-Protocol/lit-node-express/2?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#103](https://github.com/LIT-Protocol/lit-node-express/pull/103) <a href="https://app.graphite.com/github/pr/LIT-Protocol/lit-node-express/103?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#104](https://github.com/LIT-Protocol/lit-node-express/pull/104) <a href="https://app.graphite.com/github/pr/LIT-Protocol/lit-node-express/104?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.com/github/pr/LIT-Protocol/lit-node-express/2?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `next`


<h2></h2>

cl4wb0t (commented):
## Review: Does the otel-collector receive logs, metrics, and traces from lit-api-server and lit-actions?

**Short answer: No — not as currently written.** Three distinct issues prevent any OTLP data from reaching the collector.

---

### 1. lit-api-server never initializes OTLP providers

`lit-api-server/src/main.rs` calls `lit_observability::init_subscriber()` (stdout-only) and never calls `create_providers()`. The OTLP trace, metrics, and log pipelines are never started, so the collector receives nothing from lit-api-server even though the sidecar is running and wired up perfectly.

**Fix:** call `create_providers()` in `main.rs` (guarded by `#[cfg(feature = "otlp")]`) and add the resulting `TracerProvider` as a layer to the subscriber, similar to how lit-actions does it when `--enable-observability-export` is set.

---

### 2. The OTel log layer (ContextAwareOtelLogLayer) is never added to the subscriber

`init_logger_provider()` creates a `LoggerProvider`, but nothing ever adds `ContextAwareOtelLogLayer` (which bridges tracing events → OTLP logs) to the subscriber stack. This means `tracing::info!()` / `tracing::error!()` calls will **only** go to stdout, never to the OTLP log pipeline. This affects both lit-api-server (which skips create_providers entirely) and lit-actions (which calls create_providers but never wires the log layer in).

**Fix:** After `create_providers()`, add `ContextAwareOtelLogLayer::new(logger_provider)` as a layer in the subscriber chain.

---

### 3. lit-actions cannot reach the collector (network namespace issue)

The collector shares `lit-api-server`'s network namespace (`network_mode: "service:lit-api-server"`) and binds to `127.0.0.1:4317` (loopback). From inside that shared namespace, lit-api-server can reach it fine. But `lit-actions` runs in its own separate network namespace — it cannot reach loopback inside lit-api-server's namespace.

For lit-actions to push OTLP to the collector it would need to either:
- Connect to `lit-api-server:4317` (using the container hostname), **and** the collector would need to bind to `0.0.0.0:4317` instead of `127.0.0.1:4317`, or
- Join the same network namespace as lit-api-server (`network_mode: "service:lit-api-server"`), though that would need careful thought with the existing socket volume.

---

### 4. lit-actions OTLP is opt-in and disabled in docker-compose

`--enable-observability-export` defaults to `false` and the docker-compose command doesn't pass it:
```yaml
command: ["lit_actions", "--socket", "/tmp/lit_actions.sock"]
```
Even once the network issue is resolved, this flag needs to be added (or the default flipped) to actually emit OTLP.

---

### 5. No service.name resource for lit-api-server

lit-actions sets `SERVICE_NAME = "lit-actions"` on its OTel resource. lit-api-server sets no resource at all, so GCP will receive untagged telemetry and won't distinguish which service emitted what.

---

### What is working well

- The collector config covers all three signal types (traces, metrics, logs) ✓
- The GCP exporter, memory limiter, and batch processor are correctly wired ✓
- The network namespace trick for lit-api-server → collector (localhost:4317) is the right approach ✓
- The entrypoint credential bootstrap for Phala secrets is solid ✓
- The OTLP infrastructure in lit-observability (net.rs, create_providers, init_logger_provider) is correct and clean ✓

lit-api-server/Cargo.toml
```diff
@@ -74,6 +74,7 @@ sqlx = { version = "0.8", features = ["sqlite", "runtime-tokio"] }
 # rand_core = "0.6"
```
cl4wb0t, line 77: **Point 1 update:** This is actually worse than noted in the summary. `lit-api-server/Cargo.toml` defines no `otlp` feature at all — `--features otlp` would be silently ignored (or error). Before OTLP can be wired in, two things are needed:

1. Add an `otlp` feature to `lit-api-server/Cargo.toml` that enables `lit-observability/otlp`:
```toml
[features]
default = []
dstack = []
production = ["dstack"]
otlp = ["lit-observability/otlp"]
```

2. Add `#[cfg(feature = "otlp")]`-gated provider init in `main.rs`, including a `service.name` resource (see point 5), the `create_providers()` call, and `ContextAwareOtelLogLayer` on the subscriber.

otel-collector/config.yaml
```diff
@@ -0,0 +1,84 @@
...
+# from the GCP_SERVICE_ACCOUNT_JSON encrypted Phala secret).
+#
+# See: https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/googlecloudexporter
+
+receivers:
+  otlp:
+    protocols:
+      grpc:
+        # Localhost-only; the sidecar shares lit-api-server's network namespace
+        # via docker-compose network_mode: "service:lit-api-server".
```
cl4wb0t: **Network issue: lit-actions cannot reach this endpoint.**

`network_mode: "service:lit-api-server"` means `otel-collector` shares lit-api-server's network namespace and has **no hostname of its own** on the compose bridge. lit-actions is in a separate network namespace and can only reach the collector via the `lit-api-server` hostname — but this binding is loopback-only, so even `lit-api-server:4317` will be refused from lit-actions.

**Fix:** bind to `0.0.0.0` so the collector is reachable from other compose services:
```yaml
grpc:
  endpoint: 0.0.0.0:4317
http:
  endpoint: 0.0.0.0:4318
```
Then set lit-actions' OTLP endpoint to `http://lit-api-server:4317` (e.g. via `telemetry.endpoint` in its config or an env var). lit-api-server can keep using `http://127.0.0.1:4317` since it shares the namespace.

docker-compose.phala.yml
```diff
@@ -35,6 +41,25 @@ services:
       - lit-actions
     restart: unless-stopped
```
cl4wb0t, line 58: **Point 4 fix:** Add `--enable-observability-export` so lit-actions actually emits OTLP when built with `--features otlp`. Also set the endpoint to reach the collector via lit-api-server's hostname (see collector config comment):

```yaml
lit-actions:
  image: ${DOCKER_IMAGE_LIT_ACTIONS}
  command:
    - lit_actions
    - --socket
    - /tmp/lit_actions.sock
    - --enable-observability-export
  environment:
    TELEMETRY_ENDPOINT: http://lit-api-server:4317
```

And read `TELEMETRY_ENDPOINT` in `init_observability` (or map it to `telemetry.endpoint` in config) so it overrides the default `127.0.0.1:4317`.

lit-api-server/src/main.rs
```diff
@@ -17,15 +18,18 @@ use rocket_cors::{AllowedOrigins, Method};
 use rocket_okapi::okapi::openapi3::{OpenApi, Server};
```
cl4wb0t, line 14: **Points 1 & 5:** Here is the full wiring needed in `main.rs` to actually emit OTLP and set `service.name`. Add after the `set_global_default` call:

```rust
#[cfg(feature = "otlp")]
{
    use lit_observability::{
        opentelemetry::{KeyValue, global},
        opentelemetry_sdk::{Resource, propagation::TraceContextPropagator, trace as sdktrace},
        opentelemetry_semantic_conventions::resource::SERVICE_NAME,
        logging::ContextAwareOtelLogLayer,
    };

    let otel_resource = Resource::new(vec\![KeyValue::new(SERVICE_NAME, "lit-api-server")]);

    let (tracing_provider, metrics_provider, logger_provider) =
        lit_observability::create_providers(
            &lit_cfg,
            otel_resource.clone(),
            sdktrace::Config::default().with_resource(otel_resource),
        )
        .await
        .expect("Failed to create OTLP providers");

    global::set_text_map_propagator(TraceContextPropagator::new());
    global::set_tracer_provider(tracing_provider);
    global::set_meter_provider(metrics_provider);
    // Wire tracing events → OTLP log pipeline
    // (without this, tracing::info\!/error\! never reach the collector)
    tracing::subscriber::with_default(
        tracing_subscriber::registry().with(ContextAwareOtelLogLayer::new(logger_provider)),
        || {},
    );
}
```

Note: the subscriber is already set globally at this point, so the OTel log layer needs to be composed _before_ `set_global_default`, or the architecture needs a small refactor to build the full layered subscriber in one shot (including `ContextAwareOtelLogLayer`) then set it globally once.

cl4wb0t (commented):
## Review: OTLP Sidecar Observability

### Does this cover metrics AND logs? Does otel-collector receive logs from both services?

**Yes, in the current HEAD (`27fb598`), all three signal types (traces, metrics, logs) are covered for both `lit-api-server` and `lit-actions`.**

- `otel-collector/config.yaml` defines complete pipelines for `traces`, `metrics`, and `logs` — all via OTLP receiver → GCP exporter.
- Both services call `create_providers()`, set global tracer/meter providers, and wire `ContextAwareOtelLogLayer` into their subscriber stacks.
- Both services have `LIT_TELEMETRY__ENDPOINT: http://otel-collector:4317` in `docker-compose.phala.yml`, pointing to the collector on the compose bridge.

The one protection gap is Issue #7 below — `lit-api-server` has `production = ["dstack", "otlp"]` so it always emits OTLP in production images, but `lit-actions` has no such protection and relies on the image being built with `--features otlp` explicitly.

---

### Issue 1 — Missing explicit provider shutdown in `lit-api-server` (data loss on SIGTERM)

**File:** `lit-api-server/src/main.rs`

`_otlp_providers` holds `Some((metrics_provider, logger_provider))` but there is no explicit `.shutdown()` call. The batch exporters buffer spans/metrics/logs in memory and flush on shutdown — if the process receives SIGTERM, Rocket shuts down its runtime and the providers are dropped without flushing. `lit-actions` correctly calls `observability_providers.shutdown()` before exiting. `lit-api-server` should do the same.

```rust
// Before returning Ok(()) at the bottom of main():
#[cfg(feature = "otlp")]
if let Some((metrics_provider, logger_provider)) = _otlp_providers {
    let _ = metrics_provider.shutdown();
    let _ = logger_provider.shutdown();
    // tracing_provider is set globally — global::tracer_provider().shutdown()
}
```

---

### Issue 2 — `telemetry = { endpoint = "..." }` in `Rocket.toml` is a dead config key

**File:** `lit-api-server/Rocket.toml`

```toml
[default]
telemetry = { endpoint = "http://otel-collector:4317" }
```

This key is parsed by Rocket but Rocket has no knowledge of it — it will be silently ignored. The actual config is read by `LitConfig` via `cfg.get_string("telemetry.endpoint")` from its own config file chain and environment variables. The live override in production is `LIT_TELEMETRY__ENDPOINT` in `docker-compose.phala.yml`.

This entry in `Rocket.toml` will mislead developers into thinking they can change the endpoint there. It should either be removed or moved to `lit-api-server/config/default.toml` where `LitConfig` will actually read it.

---

### Issue 3 — `address = "0.0.0.0"` removed from `Rocket.toml` breaks local dev

**File:** `lit-api-server/Rocket.toml`

The old `Rocket.toml` had `address = "0.0.0.0"`. The new one omits it. Rocket's built-in default is `127.0.0.1`, so running `lit-api-server` locally outside of Docker Compose will now only bind loopback. The compose file sets `ROCKET_ADDRESS=0.0.0.0` via env var which is fine for production, but local dev without compose will silently stop being reachable from other processes/containers. Consider either keeping `address = "0.0.0.0"` in `Rocket.toml` or documenting the env var requirement in the README/PLAN.md.

---

### Issue 4 — `lit-actions` has no `production`-feature protection for OTLP

**File:** `lit-actions/cli/Cargo.toml`

`lit-api-server` is protected: its `production` feature implies `otlp`, so production images always emit OTLP. `lit-actions` has no equivalent — OTLP emission requires the image builder to explicitly pass `--features otlp`. The compose file comment documents this requirement, but there is no Dockerfile or CI enforcement visible in this PR. If CI builds `lit-actions` without `--features otlp`, the collector receives nothing from it in production and there is no error — it silently falls back to stdout. Consider adding `production = ["otlp"]` to `lit-actions/cli/Cargo.toml` to match the `lit-api-server` pattern.

---

### Issue 5 — Wrong env var name in `lit-observability/README.md`

**File:** `lit-core/lit-observability/README.md`

> `LIT_CONFIG_TELEMETRY_ENDPOINT` (defaults to `http://127.0.0.1:4317`)

The actual env var (matching LitConfig's double-underscore convention for `telemetry.endpoint`) is `LIT_TELEMETRY__ENDPOINT`. This is what `docker-compose.phala.yml` uses. The README will mislead DevOps teams trying to configure the endpoint.

---

### Issue 6 — `MetricsMiddleware` removal is a silent capability regression

**File:** `lit-core/lit-observability/src/net.rs`

The old `net.rs` contained a `MetricsMiddleware` that tracked gRPC request latency automatically. It has been removed entirely in this PR. If any gRPC service was relying on it for latency histograms, those metrics are now silently gone. This fits the "simplification" theme and may be intentional, but it should be called out explicitly in the PR description so ops/DevOps teams know to watch for missing gRPC latency metrics in their dashboards after deploy.

---

### Minor / Nits

- **`Cargo.lock` check**: Verify a build without `--features otlp` truly excludes the OTel SDK. The lock file still references `tracing-opentelemetry` — if it's not truly optional at link time, the feature gate is cosmetic.
- **`otel-collector/config.yaml`**: The `debug` exporter on the `logs` pipeline is helpful during initial rollout — make sure it's removed or set to `verbosity: basic` before a long-running production deploy to avoid log amplification.
- **`entrypoint.sh`**: The `base64 -d` call is not portable across all distros (some use `base64 --decode`). Since this runs in the otel-collector image, confirm the base image has the expected `base64` variant.

cl4wb0t (commented):
## Review: OTLP Sidecar Observability — Final State Analysis

This review covers the PR at its current HEAD (`caed5a7`). Prior review threads from `cl4wb0t` identified several issues that have since been fixed across the commit history. I'll confirm their resolution status and raise new issues not yet flagged.

---

### Resolved from earlier review threads

- **Provider shutdown in `lit-api-server`** — Fixed. `tracing_provider`, `metrics_provider`, and `logger_provider` are all explicitly shut down after `r.launch().await?`.
- **Network namespace / collector reachability** — Fixed. `otel-collector` is now a first-class service on the bridge network, not sharing `lit-api-server`'s network namespace. Both services reach it at `http://otel-collector:4317`.
- **`lit-actions` OTLP disabled** — Fixed. The `--enable-observability-export` flag was removed; OTLP is always on when the binary is built with `--features otlp`.
- **`lit-api-server` missing `service.name`** — Fixed. `Resource::new(vec![KeyValue::new(SERVICE_NAME, "lit-api-server")])` is set in `main.rs`.
- **`lit-actions` missing `production` feature** — Fixed. `lit-actions/cli/Cargo.toml` now has `production = ["otlp"]`.
- **`address = "0.0.0.0"` removed from `Rocket.toml`** — Fixed. It is back.
- **Wrong env var name in README** — Fixed. README now documents `RUST_LOG` and `LIT_TELEMETRY_ENDPOINT` correctly.

---

### Issue 1 — `lit-api-server` does not set `LIT_TELEMETRY_ENDPOINT` in `docker-compose.phala.yml`

**File:** `docker-compose.phala.yml`, `lit-api-server` service block

`lit-actions` has:
```yaml
environment:
  LIT_TELEMETRY_ENDPOINT: http://otel-collector:4317
```

`lit-api-server` has no `LIT_TELEMETRY_ENDPOINT` environment variable. `read_observability_config()` in `config.rs` falls through to the TOML file path, then defaults to `http://127.0.0.1:4317`. On the Phala CVM bridge network, `127.0.0.1` on the `lit-api-server` container is **not** where the collector is running — `otel-collector:4317` is. So `lit-api-server` will silently fail to reach the collector and fall back to stdout-only logging, despite the collector being healthy.

The `NodeConfig.toml` mount is commented out in the compose file, meaning there is no file-based override either. The only live config path is the env var.

**Fix:** Add `LIT_TELEMETRY_ENDPOINT: http://otel-collector:4317` to the `lit-api-server` environment block, matching what `lit-actions` already has.

---

### Issue 2 — `init_tonic_exporter_builder` is synchronous but creates a lazy gRPC channel

**File:** `lit-core/lit-observability/src/net.rs`

```rust
pub(crate) fn init_tonic_exporter_builder(endpoint: &str) -> Result<TonicExporterBuilder> {
    Ok(opentelemetry_otlp::new_exporter().tonic().with_endpoint(endpoint))
}
```

`create_providers` in `lib.rs` calls `init_tonic_exporter_builder` three times — once for tracing, once for metrics, once for logs. Each call creates an independent `TonicExporterBuilder` with the same endpoint. The OTLP Tonic exporter establishes its channel lazily on first export, so connection errors surface only on the first flush, not at init time. This means `create_providers` will return `Ok(...)` even if the endpoint is unreachable, and the fallback to stdout will never trigger. The OTLP init error path in both `main.rs` files is therefore dead code in the network-failure scenario. This is a tricky but real operational gap: if the collector sidecar fails to start (or has a typo in its hostname), the services will silently run with no OTLP export and no error logged.

**Recommendation:** Document this limitation in a comment near `create_providers`. Alternatively, add a health-check connection attempt after provider creation so failures surface early at startup rather than silently on first flush.

---

### Issue 3 — `tracing_provider` not registered in `ObservabilityProviders` for `lit-actions`

**File:** `lit-actions/cli/main.rs`

In the `#[cfg(feature = "otlp")]` block:

```rust
global::set_tracer_provider(tracing_provider);
// ...
ObservabilityProviders::new(metrics_provider, logger_provider)
```

The `tracing_provider` is passed to `global::set_tracer_provider` and then dropped. At shutdown, `ObservabilityProviders::shutdown()` flushes the `meter_provider` and `logger_provider`, but the global tracer provider is never explicitly shut down. Any in-flight spans buffered in the tracing pipeline will be lost on process exit. Compare this to `lit-api-server`, which correctly calls `tracing_provider.shutdown()` after `r.launch().await?`.

**Fix:** Either store `tracing_provider` in `ObservabilityProviders` and shut it down alongside the others, or call `opentelemetry::global::shutdown_tracer_provider()` (now `global::tracer_provider().shutdown()` in opentelemetry 0.x) inside `ObservabilityProviders::shutdown()`.

---

### Issue 4 — `EnvFilter` directives silently override `RUST_LOG` in counterintuitive order

**File:** `lit-core/lit-observability/src/lib.rs`

```rust
let level_filter = EnvFilter::try_from_default_env()
    .or_else(|_e| EnvFilter::from_str(log_level))
    ...
    .add_directive("hyper=error".parse().unwrap())
    .add_directive("tonic=error".parse().unwrap())
    // ...
```

`add_directive` appends to the filter **after** the base filter is built from `RUST_LOG`. Because `tracing-subscriber`'s `EnvFilter` uses last-wins semantics for overlapping directives, these appended directives will **override** whatever crate-specific level the user set in `RUST_LOG`. For example, if a developer sets `RUST_LOG=tonic=debug` to debug a gRPC issue, they will get `tonic=error` instead, with no warning. The old `setup_tracing()` in `lit-api-server` had the same pattern, so this is not a regression introduced in this PR, but it's worth calling out since the observability code is being rewritten anyway.

**Recommendation:** Document the override behavior in the `init_subscriber` doc comment, or use a `FilterFn` approach that only applies the noise-suppression directives when they are not already present in `RUST_LOG`.

---

### Issue 5 — `README.md` usage example is incorrect

**File:** `lit-core/lit-observability/README.md`

The Usage section shows:
```rust
let (subscriber, _guards) = lit_observability::init_subscriber(cfg);
```

`init_subscriber` takes `&str` (not a config object), and returns `Result<impl Subscriber + ...>` — a single value, not a tuple. The `_guards` binding does not exist in the real API. This will confuse anyone who copies the example.

**Fix:**
```rust
let subscriber = lit_observability::init_subscriber("info").expect("Failed to init subscriber");
tracing::subscriber::set_global_default(subscriber).expect("Set subscriber failed");
```

---

### Issue 6 — `tonic` and `tracing-appender` pulled in unconditionally despite being OTLP-only

**File:** `lit-core/lit-observability/Cargo.toml`

`tonic = "0.12"`, `tower = "0.5"`, `tonic-middleware = "0.2"`, and `tracing-appender = "0.2"` are unconditional dependencies. `tonic` and `tower` are only needed by `TracingMiddleware` in `net.rs::grpc`, which itself only does anything useful when `feature = "otlp"` (because `set_parent` is gated). `tracing-appender` was previously used by `simple_file_logging_subscriber`, which has been removed. These add compile-time cost and binary size for crates that only need `init_subscriber`.

This is a minor cleanup item, not a correctness issue, but worth tracking.

---

### Issue 7 — `otel-collector` image has no port bindings; port 4317 is accessible to all compose services but not to the host

**File:** `docker-compose.phala.yml`

The `otel-collector` service has no `ports:` block. This is correct for a sidecar — the collector should not be reachable from outside the CVM. However, it's worth confirming that the Phala CVM compose runner applies the same compose bridge networking model (i.e., all services on the same compose stack share a bridge network by default). If Phala's runner uses a different networking model (e.g., host networking for all services), the `otel-collector` hostname would not resolve and services would fall back to the `127.0.0.1:4317` default, which is again wrong. This is a deployment-environment question, not a code bug, but should be validated before the first production deploy.

---

### Nits

- **`otel-collector/entrypoint.sh`**: `base64 -d` works on GNU coreutils (Alpine, Debian) but some BusyBox builds only support `base64 -D`. The otel-collector-contrib image is based on a minimal Alpine variant — verify the installed `base64` binary supports `-d`. A safer portable form is `printf '%s' "$GCP_SERVICE_ACCOUNT_JSON" | base64 -d`.

- **`otel-collector/config.yaml`**: The `debug` exporter is attached to the `logs` pipeline. At the default `verbosity: basic`, this emits a one-line summary per batch to the collector's own stdout, which is fine. Just ensure this doesn't create a feedback loop if the collector's own stdout is scraped by another log collector.

- **`lit-core/lit-observability/src/logging/privacy_filter.rs`**: The `on_event` and `on_new_span` implementations are now empty bodies (comments removed). This is fine — they need to exist to satisfy the `Layer` trait — but a brief comment noting they are intentionally empty (filtering is done in `enabled`) would help future readers.

- **`Dockerfile.otel-collector`**: The `chmod +x` step requires a temporary root escalation. Consider instead using `COPY --chmod=755 otel-collector/entrypoint.sh /entrypoint.sh` (available in BuildKit / Docker 20.10+), which avoids the `USER root` layer entirely.

---

### Summary

The core architecture is sound and the most critical bugs identified in the earlier review rounds have been fixed. The single remaining correctness issue that will cause silent production breakage is **Issue 1** — `lit-api-server` missing `LIT_TELEMETRY_ENDPOINT` in compose, causing it to send OTLP to `127.0.0.1:4317` instead of the sidecar. Issue 3 (tracing provider shutdown for `lit-actions`) will cause trace data loss on clean shutdown. All other items are recommendations or documentation fixes.

otel-collector/entrypoint.sh
```diff
@@ -0,0 +1,37 @@
...
+    exit 1
+fi
+
+CREDS_FILE=/tmp/gcp-service-account.json
+
+# Support both raw JSON (starts with '{') and base64-encoded JSON.
+first_char=$(echo "$GCP_SERVICE_ACCOUNT_JSON" | head -c1)
+if [ "$first_char" = "{" ]; then
+    echo "$GCP_SERVICE_ACCOUNT_JSON" > "$CREDS_FILE"
+else
```
cl4wb0t, line 32: **Security / Correctness: credentials written via `echo` may be corrupted or leaked**

Using `echo "$GCP_SERVICE_ACCOUNT_JSON" > "$CREDS_FILE"` is problematic for two reasons:

1. **Shell word-splitting / special characters**: Service account JSON contains `{`, `}`, `:`, `"`, and newlines. Even with double-quoting the variable, `echo` on some `sh` implementations will interpret escape sequences (e.g. `\n`, `\t`) inside the value, silently corrupting the JSON. POSIX `printf '%s\n'` or a heredoc is safer.

2. **`/tmp` is world-readable on many systems**: The credentials file is written to `/tmp/gcp-service-account.json` before `chmod` is applied (there is no `chmod` here at all). Any process in the container that starts between the `>` and the `exec` can read the plaintext key. Use `install -m 600` or set `umask 077` at the top of the script before writing the file.

Suggested fixes:
```sh
umask 077
printf '%s' "$GCP_SERVICE_ACCOUNT_JSON" > "$CREDS_FILE"
# or for base64 path:
printf '%s' "$GCP_SERVICE_ACCOUNT_JSON" | base64 -d > "$CREDS_FILE"
```

otel-collector/entrypoint.sh
```diff
@@ -0,0 +1,37 @@
...
+    exit 1
+fi
+
+if [ -z "$GCP_PROJECT_ID" ]; then
+    echo "[otel-collector] ERROR: GCP_PROJECT_ID is not set" >&2
+    exit 1
+fi
+
+CREDS_FILE=/tmp/gcp-service-account.json
+
```
cl4wb0t, line 27: **Nit / Robustness: `CREDS_FILE` in `/tmp` persists across container restarts**

Because `set -e` is active and the collector user (uid 10001) owns `/tmp` inside the container, the file will linger between `docker restart` cycles. This isn't a correctness issue today, but if the secret rotates it could be stale. Consider writing to a `mktemp`-created path or explicitly removing the file on exit via a `trap`:
```sh
trap 'rm -f "$CREDS_FILE"' EXIT
```

lit-actions/cli/main.rs
```diff
@@ -19,21 +16,16 @@ struct Args {
...
 
     let args = Args::parse();
 
     let observability_rt = tokio::runtime::Runtime::new().expect("failed to create runtime");
 
-    let observability_providers = observability_rt
-        .block_on(async { init_observability(&args).await })
-        .expect("failed to init observability");
+    let observability_providers = observability_rt.block_on(async { init_observability().await });
     debug!(?args);
```
cl4wb0t, line 29: **Correctness: `init_observability` errors are silently swallowed**

Previously the call was `.expect("failed to init observability")`. Now it is:
```rust
let observability_providers = observability_rt.block_on(async { init_observability().await });
```
`init_observability` returns `ObservabilityProviders` (infallible), so individual failures inside it are handled via `eprintln!` and a fallback to stdout. That's fine for OTLP connectivity failures. However, if `lit_observability::init_subscriber` itself returns an `Err` (e.g. an invalid `RUST_LOG` string), the closure in `init_stdout` silently calls `eprintln!` and leaves the global subscriber **uninitialized**. Any subsequent `tracing::info!` / `tracing::error!` calls will be no-ops, which could mask real startup errors. Consider panicking or at least logging to stderr more prominently when the base subscriber fails to initialize, since this is a hard programmer/configuration error rather than a transient network failure.

cl4wb0t (commented):
## Overall Assessment

This is a solid, well-motivated simplification. Replacing the bespoke UDS/proxy-collector transport with a stock OTLP/gRPC sidecar significantly reduces moving parts and makes the observability stack easier to operate. The feature-gating (`otlp` / `production` Cargo features) is clean, the fallback-to-stdout path on OTLP init failure is correct, and the graceful-shutdown wiring is now more complete than before (the `TracerProvider` was previously missing from the shutdown sequence).

A handful of issues below deserve attention before merge — they range from a security/correctness problem in the entrypoint script to a silent swallow of errors, a potential uninitialized-subscriber race, and a few design choices worth discussing.

lit-api-server/src/main.rs
```diff
@@ -17,15 +17,67 @@ use rocket_cors::{AllowedOrigins, Method};
...
-use tracing_subscriber::{EnvFilter, FmtSubscriber};
 
 // The default signer count for a new instance when contracts are deployed.
 // Note that if the signers aren't funded, nothing will work until an admin sets the default api payer.
 #[rocket::main]
 #[allow(clippy::result_large_err)]
 async fn main() -> Result<(), rocket::Error> {
-    setup_tracing().expect("Failed to setup tracing.");
+    let obs = config::read_observability_config();
+
```
cl4wb0t, line 30: **Correctness: `read_observability_config()` is called before the subscriber is initialized**

`read_observability_config` silently swallows all file-read and TOML-parse errors via `.ok()`. If parsing fails for any reason (malformed TOML, permissions error) the config degrades to defaults with no indication. Because the subscriber is not yet initialized at this point in `main`, there is nowhere to log the failure. Consider returning a `Result` from `read_observability_config` and using `eprintln!` here to surface config-load issues before the subscriber exists, so operators aren't silently running with default settings when they expect overrides.

lit-api-server/src/config.rs
```diff
@@ -4,6 +4,62 @@ use std::path::Path;
...
+    // Fast path: both values from env, no file read needed.
+    if let (Some(ll), Some(ep)) = (log_level.as_ref(), telemetry_endpoint.as_ref()) {
+        return ObservabilityConfig {
+            log_level: ll.clone(),
+            telemetry_endpoint: ep.clone(),
+        };
+    }
+
+    let toml_path = Path::new("NodeConfig.toml");
+    let doc = if toml_path.exists() {
```
cl4wb0t: **Design: relative path `NodeConfig.toml` is CWD-dependent**

`Path::new("NodeConfig.toml")` resolves relative to the process's current working directory at startup. In the Docker container the CWD is `/app`, which matches the volume-mount path in `docker-compose.phala.yml`, so this works today. But it's a fragile implicit contract — if someone runs the binary from a different directory in testing or a future deployment, the config file will be silently skipped and defaults used. Consider accepting the path via an environment variable (e.g. `NODE_CONFIG_PATH`) with a fallback to the current hardcoded relative path, or documenting the CWD requirement explicitly.

lit-api-server/src/config.rs
```diff
@@ -4,6 +4,62 @@ use std::path::Path;
...
+
+/// Reads observability settings from NodeConfig.toml if present, otherwise returns defaults.
+/// These can be overridden at runtime via `RUST_LOG` (log level) and
+/// `LIT_TELEMETRY_ENDPOINT` (collector endpoint).
+pub fn read_observability_config() -> ObservabilityConfig {
+    let log_level = env::var("RUST_LOG").ok();
+    let telemetry_endpoint = env::var("LIT_TELEMETRY_ENDPOINT").ok();
+
+    // Fast path: both values from env, no file read needed.
+    if let (Some(ll), Some(ep)) = (log_level.as_ref(), telemetry_endpoint.as_ref()) {
```
cl4wb0t: **Logic: "fast path" only triggers when *both* env vars are set**

The comment says "Fast path: both values from env, no file read needed." This is correct but perhaps surprising: if `RUST_LOG` is set but `LIT_TELEMETRY_ENDPOINT` is not, the code still reads and parses the TOML file, even though the `RUST_LOG` value from env will be used and the only thing the file can contribute is `telemetry_endpoint`. This is a minor efficiency concern but also means that a malformed `NodeConfig.toml` will cause a silent failure even when the operator has explicitly set `RUST_LOG` to override it. Document this behavior or restructure so that each value independently short-circuits.

lit-core/lit-observability/Cargo.toml
```diff
@@ -9,29 +9,27 @@ edition.workspace = true
 default = []
 testing = ["ansi", "tracing-log"]
 ansi = []
-tracing-log = []
-proxy-collector = []
-channels = ["dep:flume"]
+tracing-log = ["dep:tracing-log"]
+channels = ["dep:flume", "otlp"]
+otlp = ["dep:opentelemetry-otlp", "dep:opentelemetry_sdk", "dep:tracing-opentelemetry"]
```
cl4wb0t, line 14: **Dependency: `channels` feature unconditionally pulls in `otlp`**

```toml
channels = ["dep:flume", "otlp"]
```

The `channels` feature (used by `lit-api-server` as a non-OTLP build) now implies `otlp`, which means any consumer that enables `channels` without wanting OTLP will silently pull in `opentelemetry-otlp`, `opentelemetry_sdk`, and `tracing-opentelemetry`. In `lit-api-server/Cargo.toml`, the crate is declared as:
```toml
lit-observability = { path = "../lit-core/lit-observability", features = ["channels"] }
```
This means even a non-`otlp` build of `lit-api-server` will compile in all the OTLP deps. If the intent is that `channels` always requires OTLP, that should be documented. If not, `channels` and `otlp` should remain independent features.

lit-core/lit-observability/Cargo.toml
```diff
@@ -9,29 +9,27 @@ edition.workspace = true
...
 opentelemetry.workspace = true
-opentelemetry-otlp = { workspace = true, features = ["logs"] }
+opentelemetry-otlp = { workspace = true, features = ["logs"], optional = true }
 opentelemetry-semantic-conventions.workspace = true
-opentelemetry_sdk = { workspace = true, features = ["logs"] }
+opentelemetry_sdk = { workspace = true, features = ["logs"], optional = true }
 tokio.workspace = true
 tonic = "0.12"
 tonic-middleware = "0.2"
 tower = "0.5"
```
cl4wb0t, line 28: **Dependency: `tonic` and `tonic-middleware` are unconditional**

`tonic = "0.12"` and `tonic-middleware = "0.2"` remain as unconditional (non-optional) dependencies even though they are only used by the OTLP export path and the `grpc` module. This adds non-trivial compile time to builds that don't need OTLP. Consider making them optional under `otlp` (or a separate `grpc` feature) if compile-time budget is a concern.

otel-collector/config.yaml
```diff
@@ -0,0 +1,84 @@
...
+
+receivers:
+  otlp:
+    protocols:
+      grpc:
+        # All interfaces so lit-api-server and lit-actions can both reach the
+        # collector via the otel-collector hostname on the compose bridge network.
+        endpoint: 0.0.0.0:4317
+      http:
+        endpoint: 0.0.0.0:4318
```
cl4wb0t: **Security: HTTP receiver (port 4318) is enabled but not used and not firewalled**

The config exposes an HTTP OTLP receiver on `0.0.0.0:4318`, but the application services only send to gRPC `4317`. The HTTP receiver is unused and adds unnecessary attack surface on the compose bridge network. Remove it or restrict to `127.0.0.1:4318` if it is not intentionally required.

otel-collector/config.yaml
```diff
@@ -0,0 +1,84 @@
...
+
+processors:
+  memory_limiter:
+    check_interval: 1s
+    limit_mib: 256
+    spike_limit_mib: 64
+
+  batch:
+    send_batch_size: 512
+    timeout: 5s
```
cl4wb0t, line 29: **Performance: `memory_limiter` should appear first in every pipeline**

The OTel Collector docs recommend placing `memory_limiter` as the **first** processor in every pipeline so it can drop data before downstream processors (like `batch`) allocate additional memory. Currently the pipelines use `[memory_limiter, resource, batch]` which is correct ordering — good. However, the `batch` processor's `send_batch_size: 512` with `timeout: 5s` can hold up to 512 * (average span size) in memory before flushing. On a resource-constrained Phala CVM with `limit_mib: 256` this could cause `memory_limiter` to kick in frequently. Consider also setting `send_batch_max_size` to cap memory per batch.

docker-compose.phala.yml
```diff
@@ -29,10 +45,30 @@ services:
...
+    restart: unless-stopped
+
+  # OTel Collector — receives OTLP (gRPC/HTTP) on 0.0.0.0:4317/4318 from all
+  # services on the compose bridge, then exports to GCP Cloud Monitoring,
+  # Cloud Trace, and Cloud Logging.
+  #
+  # NOTE: both lit-api-server and lit-actions must be built with --features otlp
+  # for OTLP emission. Without that feature they fall back to stdout logging.
+  otel-collector:
+    image: ${DOCKER_IMAGE_OTEL_COLLECTOR}
```
cl4wb0t, line 67: **Reliability: `otel-collector` has no `healthcheck` defined**

Both `lit-api-server` and `lit-actions` depend on `otel-collector` with `condition: service_started`, which only waits for the container to start — not for the gRPC listener to be ready. If the collector takes a moment to bootstrap (e.g. while `entrypoint.sh` decodes credentials), the application services may attempt OTLP connections before the collector is listening, relying on the SDK's internal retry logic. Adding a healthcheck (e.g. `grpc_health_probe` or a simple `nc -z localhost 4317`) and changing to `condition: service_healthy` would make startup sequencing deterministic.

lit-api-server/src/main.rs
```diff
@@ -138,21 +174,20 @@ async fn main() -> Result<(), rocket::Error> {
...
-        .add_directive("ethers_providers::rpc::provider=warn".parse()?); // listeners
-
-    let subscriber = FmtSubscriber::builder()
-        .with_max_level(Level::TRACE)
-        .with_env_filter(env_filter)
-        .finish();
-    tracing::subscriber::set_global_default(subscriber).expect("setting default subscriber failed");
+    #[cfg(feature = "otlp")]
+    if let Some((tracing_provider, metrics_provider, logger_provider)) = _otlp_providers {
+        if let Err(e) = tracing_provider.shutdown() {
```
cl4wb0t, line 179: **Correctness: shutdown uses `eprintln!` after the subscriber is torn down**

By the time this shutdown block runs, the tracing subscriber may already be in an indeterminate state (Rocket has returned from `launch()`). Using `eprintln!` here is intentional and correct — good. However, `tracing_provider.shutdown()` flushes pending spans synchronously. If the Tokio runtime has already begun teardown, this blocking call may hang or panic. Consider wrapping each shutdown in a timeout (e.g. using `std::time::Duration`) or confirming that Rocket's `#[rocket::main]` guarantees the runtime is still alive after `launch()` returns.

lit-api-server/src/main.rs
```diff
@@ -139,21 +183,20 @@ async fn main() -> Result<(), rocket::Error> {
...
+        }
+        if let Err(e) = metrics_provider.shutdown() {
+            eprintln!("Failed to shutdown metrics provider: {e}");
+        }
+        if let Err(e) = logger_provider.shutdown() {
+            eprintln!("Failed to shutdown logger provider: {e}");
+        }
+    }
+
     Ok(())
```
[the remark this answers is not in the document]
Garandor: fixed

GTC6244 (approved):

PR #5, opened 2026-02-24
use self-hosted runner and fix workflow

Garandor:
### TL;DR

Added dynamic runner selection with fallback capability and switched Docker username from secret to variable in the Phala deployment workflow.

### What changed?

- Added a new `determine-runner` job that uses `mikehardy/runner-fallback-action@v1` to select between self-hosted runners and ubuntu-latest as fallback
- Modified the `deploy` job to depend on the runner determination and use the selected runner type
- Changed Docker login to use `vars.DOCKERHUB_USERNAME` instead of `secrets.DOCKERHUB_USERNAME`

### How to test?

1. Trigger the workflow manually via workflow_dispatch or through the configured triggers
2. Verify that the runner selection works correctly by checking which runner type is used in the job logs
3. Confirm that Docker login still works with the username now sourced from repository variables
4. Test both scenarios where self-hosted runners are available and unavailable to ensure fallback works

### Why make this change?

This change improves workflow reliability by automatically falling back to GitHub-hosted runners when self-hosted runners are unavailable, preventing deployment failures due to runner unavailability. Moving the Docker username to variables also follows better practices for non-sensitive configuration data.


Garandor:
* **#5** <a href="https://app.graphite.com/github/pr/LIT-Protocol/lit-node-express/5?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/LIT-Protocol/lit-node-express/5?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #28, opened 2026-02-28
chore(justfile): split into .just/ sub-files by theme

cl4wb0t:
## Summary

- Moves all recipes out of the root `justfile` into three focused files under `.just/`:
  - **`.just/build.just`** — `build`, `actions`, `api-server`, `clean`
  - **`.just/deploy.just`** — `setup`, `docker-build`, `docker-push`, `deploy`, `docker-run-local`, `ssh`, `_check_docker`, `_check_phala`
  - **`.just/test.just`** — `test` (k6)
- The root `justfile` now contains only shared variables (`image`, `app_name`, `instance_type`), the `default` recipe, and `import` directives.
- Recipe names, behaviour, and `just --list` output are **unchanged**.

## Test plan

- [ ] `just --list` produces the same recipe list as before
- [ ] Spot-check: `just build`, `just test smoke`, `just deploy --dry-run` (or equivalent) resolve correctly

### DIFF
--- a/.just/build.just
+++ b/.just/build.just
@@ -1,0 +1,14 @@
+[group: 'build']
+build: actions api-server
+
+[group: 'build']
+actions:
+    cargo build --manifest-path=lit-actions/Cargo.toml --bin lit_actions --all-features
+
+[group: 'build']
+api-server:
+    cargo build --manifest-path=lit-api-server/Cargo.toml --bin lit-api-server --all-features
+
+clean:
+    cargo clean --manifest-path=lit-actions/Cargo.toml
+    cargo clean --manifest-path=lit-api-server/Cargo.toml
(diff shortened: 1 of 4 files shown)


clawdbot-glitch003:
Closing as no longer relevant. Since this was opened (Feb 28), `main` has already modularized the justfile via `import "justfile.deploy"` and `import "justfile.sim"`, so the reorg goal is already met (just with a `justfile.*` naming convention rather than a `.just/` directory). The recipe content has also diverged substantially — the deploy recipes were rewritten and moved to `justfile.deploy` (now digest-pinned), the single `image` var was replaced by three (`image_lit_actions`/`image_lit_api_server`/`image_otel_collector`), new vars and recipes were added (`fmt`, `clippy`, `contracts-*`), and `test` gained an `ecdsa-sign` case + correlation IDs. The conflicts reflect edits to recipes that no longer exist in this form, so this would need a from-scratch redo. If we specifically want everything under `.just/` instead of the current `justfile.*` siblings, that's better as a fresh small PR.

PR #36, opened 2026-03-02
Feature/cpl 23 add a readonly version of getting a contract

GTC6244:
### WHAT

And a standard HTTP provider to read contracts for cases when we don't need mutable state providers.


GTC6244:
weird extra code in here!

PR #49, opened 2026-03-03
fix: use dstack-verifier Docker image for full attestation validation

Garandor:
## Summary
- The bare `dstack-verifier` binary fails `is_valid` because it shells out to `dstack-acpi-tables` (a custom QEMU build) for RTMR0 measurement computation, and that binary isn't available on the runner
- Switches to building and caching the full verifier Docker image from `verifier/builder/Dockerfile`, which bundles `dstack-acpi-tables`, QEMU BIOS files, `tar`, `sha256sum`, and shared libraries
- Docker image is cached by dstack HEAD SHA, so rebuilds only happen when upstream changes

## Test plan
- [ ] Verify the Docker image builds successfully on the self-hosted runner
- [ ] Confirm `is_valid=true` for the deployed `next` CVM attestation
- [ ] Confirm cache hit skips the Docker build on subsequent runs

### DIFF
--- a/.github/workflows/verify-attestation.yml
+++ b/.github/workflows/verify-attestation.yml
@@ -1,10 +1,9 @@
 # Verify attestation from a deployed Phala CVM
 #
-# Fetches /attestation from the live API, runs dstack-verifier (via its
-# Docker image which bundles dstack-acpi-tables and all dependencies),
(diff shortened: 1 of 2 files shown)


Garandor:
done in https://github.com/LIT-Protocol/lit-node-express/pull/75

PR #62, opened 2026-03-03
Add GetKey for Dstack integration

GTC6244:
### WHAT

I replaced the standard random wallet creation with a creation from Dstack, though the seeding is still pseudo-random right now ( and may stay that way? ) .


Garandor (approved):
LGTM!

PR #68, opened 2026-03-05
Deterministic API Payer

GTC6244:
### WHAT

PR that takes a first pass at using deterministic signers so that we can remove secrets from our VMs and use the Phalla ( or other KMS ) to generate signers that get used for the dashboard through the API.


glitch003 (approved):
lgtm!

PR #69, opened 2026-03-06
Concurrency - Mutable Contract Calls through the API 

GTC6244:
## WHAT

Creates a background thread with a pool of signers to allow for concurrent signing when using the API. 

Adds concurrent contract signing support in lit-api-server via a background signer pool, updates AccountConfig contract/admin tooling to support an owner + setApiPayer, and extends the Lit Admin UI and Lit Actions JS SDK with new convenience functionality.

Changes:

- Introduces a SignerPool background task and threads it through core account-management endpoints to support concurrent signing.
- Updates AccountConfig Solidity facets + generated JSON artifact to add owner() and setApiPayer(address) guarded by api_payer or owner.
- Extends the Lit Admin dapp UI to show owner, api payer balance, fetch node chain config / api payers, and set api_payer

### Codex reviews to ignore:

_- [P1] Guard against zero-size signer pool initialization — /Users/brendonpaul/GitHub/lit-node-express/lit-api-server/src/accounts/signer_pool.rs:99-99
  When `apiPayerCount()` is `0` (which is the default right after deployment), `start_signer_pool` passes `pool_size = 0` into `get_signer_entries`, and `start_index + pool_size - 1` underflows in the range expression. That can panic at startup (or produce an invalid huge range), preventing the node from booting and blocking the intended API payer bootstrap flow._

While this sounds correct, it actually functions properly as designed - starting with a pool of 0 signers does NOT cause any issues.  ( start_index =1, during initilization, so the underflow never occurs ). 

_- [P2] Export newly added LitActions helpers — /Users/brendonpaul/GitHub/lit-node-express/lit-actions/ext/js/02_litActionsSDK.js:137-155
  This patch adds `getRpcUrl` and `encrypt_bls`, but they are not added to `globalThis.LitActions`. As a result, callers cannot access `Lit.Actions.getRpcUrl` or `Lit.Actions.encrypt_bls` at runtime, so the new SDK surface is effectively non-functional._

Sounds like this comes from a bad merge somewhere else.  Will address in a separate PR, as getRpcUrl and encrypt_Bls functions were deprecated.


lit-api-server/src/accounts/signer_pool.rs
```diff
@@ -0,0 +1,172 @@
...
+                            if let Some(entry) = entries
+                                .iter_mut()
+                                .min_by_key(|e| e.in_use_since.unwrap_or(Instant::now()))
+                            {
+                                let _ = reply.send(SignerHandle {
+                                    client: Arc::clone(&entry.client),
+                                    address: entry.address,
+                                });
+                            }
+                        }
```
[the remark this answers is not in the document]
GTC6244: intentional.... though it might make sense to extend/contract instead, since the signers are deterministic.

lit-api-server/src/core/account_management.rs
```diff
@@ -142,7 +150,7 @@ pub async fn create_wallet(api_key: &str) -> Result<CreateWalletResponse, ApiSta
     .await?;
 
     Ok(CreateWalletResponse {
-        wallet_address: bytes_to_hex(wallet_address.as_bytes()),
+        wallet_address: bytes_to_0x_hex(wallet_address.as_bytes()),
     })
 }
```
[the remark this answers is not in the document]
GTC6244: breaking is ok.

lit-api-server/src/accounts/signable_contract.rs
```diff
@@ -94,3 +76,35 @@ pub(crate) async fn get_read_only_account_config_contract()
...
+            >,
+        >,
+        NonceManagerMiddleware<
+            SignerMiddleware<Provider<ethers_providers::Http>, Wallet<SigningKey>>,
+        >,
+        (),
+    >,
+    signer_pool: Arc<SignerPool>,
+    signer_address: H160,
+) -> Result<bool> {
```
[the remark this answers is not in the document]
GTC6244: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/LIT-Protocol/lit-node-express/pull/69#discussion_r2893745774)

lit-static/static/dapps/litAdmin/app.js
```diff
@@ -110,3 +138,164 @@ async function fetchContractValues() {
...
+      listEl.innerHTML = '<span style="color:var(--muted);font-family:\'JetBrains Mono\',monospace;font-size:0.85rem">No payers returned.</span>';
+      return;
+    }
+
+    listEl.innerHTML = payers.map((addr, i) =>
+      `<div class="result-row">` +
+        `<span class="result-label">Payer ${i + 1}</span>` +
+        `<span class="result-value">${addr}</span>` +
+      `</div>`
+    ).join('');
```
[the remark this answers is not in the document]
GTC6244: ignore - temp app.

lit-static/static/dapps/litAdmin/app.js
```diff
@@ -110,3 +138,164 @@ async function fetchContractValues() {
...
+  const btn = el('btn-set-api-payer');
+
+  if (!contractAddress) {
+    showWriteStatus('Enter the contract address above first.', true);
+    return;
+  }
+  if (!ethers.isAddress(newApiPayer)) {
+    showWriteStatus('Enter a valid Ethereum address for the new api_payer.', true);
+    return;
+  }
```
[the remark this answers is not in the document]
GTC6244: demo app.  not concerned.

lit-static/static/dapps/litAdmin/app.js
```diff
@@ -110,3 +138,164 @@ async function fetchContractValues() {
...
+    const provider = await connectWallet();
+console.log("provider", provider);
+    if (!provider) return;
+
+    
+    const signer = await provider.getSigner();
+    console.log("signer:", signer);
+    const contract = new ethers.Contract(contractAddress, SET_API_PAYER_ABI, signer);
+    console.log("contract:", contract);
+
```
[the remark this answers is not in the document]
GTC6244: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/LIT-Protocol/lit-node-express/pull/69#discussion_r2893745793).   Remove console.log statements in this case.

k6/k6-script.sample.ts
```diff
@@ -1,9 +1,6 @@
-import { checkAndLog } from "./check.ts";
 import { LitApiServerClient } from "./litApiServer.ts";
 
-const baseUrl =
-  __ENV.BASE_URL ||
-  "https://e364da71b0c9af3b9068daa6321edd6ee932aa89-8000.dstack-pha-prod5.phala.network/core/v1";
+const baseUrl = "<BASE_URL>";
```
Garandor, line 3: @GTC6244 did you regenerate this? this diff seems unexpected

Garandor: actually, nvm this is fine - we don't use this sample script for anything anyway

k6/openapi.json
```diff
@@ -0,0 +1,1028 @@
+{
```
Garandor, line 1: i dont think this file should be checked in.
it is normally read from the live system and used to generate apiserver and sample.ts

GTC6244: No?   Doesn't that mean that if we change something tests will fail ?    

In PR #76 , I updated the tests to reflect changes in this PR.   If I'm using the live system as reference for generation, then these tests are only applicable to the live system, and never useable in CI.   Or is that the point?    And if so, the rule is to never update these tests until prod has been updated.

glitch003 (approved):
lgtm!

k6/openapi.json
Garandor: There is a bit of a catch-22 there in that you can't update the tests without an already deployed system, so maybe we _should_ check it in.
How did you generate the openapi.json @GTC6244 ?

Garandor: Originally, i wanted these tests to run against the deployed system only - i.e. not in CI.
There is no reason _not_ to also run it in CI against a simulated env with anvil and dstack-sim though - since we're running on testnet and self-hosted hardware anyway and not burning real tokens or CI time.

If that is what this change allows, proceed :)

Garandor: I suppose the CI job that executes those doesnt exist yet so it's a preparation for later

GTC6244: @Garandor  - I actually just copied the file.   Figured that since I was manually running a tool, I could manually copy - however, I would suggest that this is NOT a great practice.   I think I saw something about being able to run the binary in a mode that just spits out out for documentation purposes - I'll take a look.

PR #79, opened 2026-03-09
Feature/cpl 62 update k6 tests (2)

GTC6244:
### WHAT

Updated to use newer versions of contract and API.


Garandor:
Merging, only comment changes left after merge of https://github.com/LIT-Protocol/lit-node-express/pull/83

PR #103, opened 2026-03-12
add fairings to e2e trace from request to response

Garandor:


Garandor:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/LIT-Protocol/lit-node-express/103?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#103** <a href="https://app.graphite.com/github/pr/LIT-Protocol/lit-node-express/103?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/LIT-Protocol/lit-node-express/103?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#2** <a href="https://app.graphite.com/github/pr/LIT-Protocol/lit-node-express/2?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#104](https://github.com/LIT-Protocol/lit-node-express/pull/104) <a href="https://app.graphite.com/github/pr/LIT-Protocol/lit-node-express/104?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `next`


<h2></h2>

cl4wb0t (commented):
## Requirements Verification: End-to-End Tracing

Reviewing against `architectureDocs/deployment/planning/observability/requirements.md`.

---

### Initialization Requirements

| Requirement | Status | Notes |
|---|---|---|
| stdout/err logging if cannot terminate | **PASS** | `init_subscriber` provides fmt layer to stdout. Both `lit-api-server` and `lit-actions` fall back to stdout if OTLP fails. |
| OTel export optional, collector path required or terminate | **PASS** | `lit-api-server` requires `LIT_TELEMETRY_ENDPOINT` when `otlp` feature enabled, exits on missing. `lit-actions` does the same. |
| If path provided but unreachable, log error prominently | **PASS** | Both crates log `eprintln!("OTLP init failed ...")` and fall back to stdout. |
| Export logs, metrics, and traces to OTel Collector → GCP | **PASS** | `create_providers` initializes all three pipelines. Collector config exports to `googlecloud`. |

### Data Requirements

| Requirement | Status | Notes |
|---|---|---|
| End-to-end traceable from request to response | **PASS** | `ObservabilityFairing` creates a span per request with `request_id` + `correlation_id`. |
| Random span ID created on request arrival | **PASS** | `Uuid::new_v4()` generated in `on_request`. |
| X-Request-Id on response (success AND failure) | **PASS** | `on_response` attaches it unconditionally. Tests cover 200, 500, and 404. |
| User-provided X-Request-Id MUST be ignored | **PASS** | Fairing always generates a new UUID, never reads incoming X-Request-Id. Test `user_provided_x_request_id_ignored_new_uuid_created` verifies. |
| X-Correlation-Id propagated if user provides it | **PASS** | Extracted in `on_request`, returned in `on_response` only when present. |
| X-Correlation-Id MUST NOT be on response if user didn't provide it | **PASS** | `on_response` guards with `if let Some(ref id)`. Two tests verify absence. |

### Test Coverage

Excellent. 8 focused tests covering every MUST/MUST NOT from the requirements, including edge cases (404 via catcher, empty header, user-sent X-Request-Id ignored).

---

## Issues Found

### 1. **BUG: Request span is dropped before the request is actually processed**

`fairing.rs:124-131` — The span is created and entered via `let _span = info_span!(...).entered()`, but `_span` is dropped at the end of `on_request`. This means the span covers only the fairing execution, NOT the actual route handler. The "end-to-end traceable from request to response" requirement implies the span should cover the full request lifecycle. Since Rocket fairings don't support wrapping the entire request lifecycle in a span (unlike middleware in tower/axum), the span here is ephemeral. The `set_request_context` call does propagate IDs to logs via task-local/span-extension storage, so log correlation works, but the OTel *trace span* will show only the fairing duration, not the full request.

**Suggestion**: Document this limitation explicitly, or consider using Rocket's `AdHoc` fairing with a request-scoped guard that holds the span guard, or accept the fairing-scoped span and document that trace duration doesn't reflect full request time.

### 2. **BUG: `clear_task_request_context` is never called**

`set_request_context` is called in `on_request` (line 133), which inserts into the global `TASK_CONTEXTS` `DashMap`. But `clear_task_request_context()` is never called in `on_response` or anywhere in the request lifecycle. Under Rocket's async model, tokio tasks may be reused across requests. If a subsequent request happens to land on the same tokio task ID, it will inherit stale context from a previous request until `set_request_context` overwrites it. More importantly, the `DashMap` entries accumulate without bound — this is a **memory leak** proportional to the number of unique tokio task IDs over the server's lifetime.

**Fix**: Call `lit_observability::logging::clear_task_request_context()` at the end of `on_response`.

### 3. **Correlation ID not propagated to lit-actions gRPC calls**

The `CorrelationId` request guard extracts the user's correlation ID and passes it to `core_features::lit_action()`. But in `core_features.rs:40`, `let request_id = correlation_id.clone()` passes the *correlation* ID as the `request_id` to `ClientBuilder`. Then in `client.rs:65`, only `x-request-id` is set on the gRPC metadata — the `x-correlation-id` is never propagated to lit-actions.

This means:
- The lit-actions `TracingMiddleware` in `net.rs` receives `x-request-id` (which is actually the correlation ID, not the UUID generated by the fairing)
- The fairing-generated `request_id` (the UUID) is lost in the gRPC hop
- If the user didn't provide a correlation ID, `correlation_id.0` is `None`, so `request_id` in `ClientBuilder` is `None`, and the gRPC metadata `x-request-id` will fail to parse

**Fix**: Pass both `request_id` (from `RequestTracingContext`) and `correlation_id` through to the gRPC metadata. Add a `RequestId` request guard similar to `CorrelationId`.

### 4. **gRPC `TracingMiddleware` (net.rs) always sets X-Correlation-Id on response**

`net.rs:84-88` — The gRPC middleware always attaches `x-correlation-id` to the response, even when the caller didn't provide one (it falls back to a new UUID or the request_id). The requirements state: *"X-Correlation-Id MUST NOT be attached as header on a response if the user did not provide a value for it."* The HTTP fairing correctly implements this conditional, but the gRPC middleware does not.

### 5. **`CorrelationId` OpenAPI description is misleading**

`fairing.rs:82` — The OpenAPI description says "Auto-generated if not provided" but the requirements say the opposite: correlation ID is only echoed back if the user provides it. It is not auto-generated. This could mislead API consumers into expecting a correlation ID on every response.

### 6. **Security: entrypoint.sh credential handling**

`otel-collector/entrypoint.sh:29` — `echo "$GCP_SERVICE_ACCOUNT_JSON" > "$CREDS_FILE"` writes credentials to `/tmp` with default permissions (world-readable in many base images). Add `umask 077` before the write, and use `printf '%s' "$var"` instead of `echo` to avoid shell interpretation of escape sequences in the JSON.

---

## Summary

The implementation faithfully covers all the HTTP-layer requirements with good test coverage. The main gaps are:

1. **Request span lifetime** doesn't cover the full request (architectural limitation of Rocket fairings)
2. **Memory leak** from never clearing task-local context
3. **Broken context propagation** to lit-actions — correlation ID and request ID are confused/lost at the gRPC boundary
4. **gRPC middleware** violates the X-Correlation-Id conditional requirement

PR #108, opened 2026-03-12
Feature/cpl 78 check post vs get endpoints

GTC6244:
### WHAT

Check for post vs get endpoints.   It was really just one that was of concern - the getIPFScid endpoint.


GTC6244:
merged in another branch.

PR #109, opened 2026-03-12
ci: fail if contract bindings are out of date

cl4wb0t:
## Summary
- Adds a `contract-bindings` job to the Rust CI workflow
- Runs `make generate` in `lit-api-server/blockchain/lit_node_express/` and fails if the generated Rust bindings in `src/accounts/contracts/` differ from what is committed

## Test plan
- [ ] Verify the new job runs successfully on a branch where bindings are up to date
- [ ] Verify the job fails when bindings are stale (e.g. modify a generated file and push)


Garandor (approved):
LGTM

PR #119, opened 2026-03-12
Better stringify handling, lit-actions on it's own ui page

GTC6244:


GTC6244:
merged in a different PR.

PR #121, opened 2026-03-13
fix: extend http_request span to cover full request lifecycle

cl4wb0t:
## Summary
- The `http_request` `info_span` was `.entered()` and dropped at the end of `on_request`, giving it near-zero duration in traces
- Now the `Span` (which is `Send + Sync`, unlike `EnteredSpan`) is stored in Rocket's `local_cache`, so it lives from `on_request` through `on_response`
- The OTel layer records span duration from creation to drop, producing a trace span that covers actual request handling time
- The span is briefly entered during `on_request` to set request context (via `Span::current()`), then stored un-entered for the rest of its lifetime

Depends on #120 (adds the `OpenTelemetryLayer` that exports these spans).

## Test plan
- [ ] Deploy with #120 merged and verify `http_request` spans in GCP Cloud Trace have meaningful duration (not near-zero)
- [ ] Verify `request_id` and `correlation_id` attributes appear on the span
- [ ] All fairing tests pass locally

### DIFF
--- a/lit-actions/cli/main.rs
+++ b/lit-actions/cli/main.rs
@@ -107,8 +107,8 @@ async fn init_observability() -> ObservabilityProviders {
         global::set_meter_provider(metrics_provider.clone());
 
         let tracer = tracing_provider.tracer("lit-actions");
-        let otel_trace_layer = lit_observability::tracing_opentelemetry::layer()
-            .with_tracer(tracer);
+        let otel_trace_layer =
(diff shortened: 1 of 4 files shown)


cl4wb0t:
Superseded by #122, which includes this span-lifecycle fix plus RequestSpan guard, http.status_code recording, request_id/correlation_id propagation to lit-actions, OTel trace context injection, and privacy mode tagging.

PR #132, opened 2026-03-13
feat: CVM system observability via hostmetrics receiver

cl4wb0t:
## Summary
- Configures the **OTel Collector** `hostmetrics` receiver to collect CVM-level CPU, memory, disk I/O, network, and load metrics directly — no extra sidecar needed
- Mounts only `/proc` and `/sys` (read-only) into the collector container to minimise the host read surface
- Wires scraped metrics into the existing `metrics` pipeline so system stats flow to **GCP Cloud Monitoring** alongside existing application telemetry

## Changes
| File | What |
|---|---|
| `docker-compose.phala.yml` | `pid: host` and targeted `/proc`, `/sys` bind-mounts on the otel-collector service |
| `otel-collector/config.yaml` | `hostmetrics` receiver with cpu, memory, disk, network, load scrapers; disk excludes virtual devices; metrics pipeline updated to `[otlp, hostmetrics]` |

## Design decisions
- **No node-exporter sidecar** — otel-collector-contrib ships the `hostmetrics` receiver, so we collect metrics directly without an extra container
- **No full root mount** — only `/proc` and `/sys` are mounted to limit the read surface in the CVM. This means the `filesystem` scraper (disk *space*) is omitted since it requires `statfs()` on real mount points. Disk I/O throughput is still reported via `/proc/diskstats`
- **Disk device excludes** — `loop.*`, `fd.*`, `sr.*` are filtered out to avoid noisy virtual-device metrics

## Metrics available after merge
- `system.cpu.time` — per-CPU mode time (user, system, idle, iowait, etc.)
- `system.cpu.load_average.1m`, `5m`, `15m` — system load averages
- `system.memory.usage` — RAM usage by state (used, free, buffered, cached, etc.)
- `system.disk.io`, `system.disk.operations` — disk I/O throughput and ops
- `system.network.io`, `system.network.packets` — network I/O

## Test plan
- [ ] Verify the `next` branch deploy-phala workflow succeeds
- [ ] Confirm otel-collector starts without errors related to the hostmetrics receiver
- [ ] Validate metrics appear in GCP Cloud Monitoring (Metrics Explorer → `system.cpu.time`)
- [ ] Check OTel Collector logs for successful hostmetrics scrape cycles (no errors)

### DIFF
--- a/docker-compose.phala.yml
+++ b/docker-compose.phala.yml
@@ -84,11 +84,13 @@ services:
       # Phala CVM secrets in the deployment manifest.
       GCP_SERVICE_ACCOUNT_JSON: ${GCP_SERVICE_ACCOUNT_JSON}
       GCP_PROJECT_ID: ${GCP_PROJECT_ID}
-    # Host filesystem + pid namespace for the hostmetrics receiver to report
-    # accurate CVM-level CPU, memory, disk, and network metrics.
+    # Host pid namespace + /proc and /sys mounts for the hostmetrics receiver
+    # to report accurate CVM-level CPU, memory, disk, and network metrics.
+    # We avoid mounting the entire root filesystem to limit the read surface.
     pid: host
     volumes:
-      - '/:/host:ro,rslave'
+      - '/proc:/host/proc:ro'
+      - '/sys:/host/sys:ro'
     restart: unless-stopped
 
   lit-static:
--- a/docker-compose.phala.yml
+++ b/docker-compose.phala.yml
@@ -97,5 +97,19 @@ services:
       ROCKET_PORT: "8001"
     restart: unless-stopped
 
+  # Prometheus Node Exporter — exposes CVM host metrics (CPU, memory, disk, network)
+  # on port 9100 for the OTel Collector to scrape. The /host bind-mount + --path.rootfs
+  # gives accurate host-level readings from inside the container. pid:host is required
+  # for per-process and CPU metrics.
+  # See: https://docs.phala.com/phala-cloud/production-checklist
+  node-exporter:
+    image: quay.io/prometheus/node-exporter:v1.9.0
+    command:
+      - '--path.rootfs=/host'
+    pid: host
+    volumes:
+      - '/:/host:ro,rslave'
+    restart: unless-stopped
+
 volumes:
   lit-socket:
--- a/docker-compose.phala.yml
+++ b/docker-compose.phala.yml
@@ -84,6 +84,11 @@ services:
       # Phala CVM secrets in the deployment manifest.
       GCP_SERVICE_ACCOUNT_JSON: ${GCP_SERVICE_ACCOUNT_JSON}
       GCP_PROJECT_ID: ${GCP_PROJECT_ID}
+    # Host filesystem + pid namespace for the hostmetrics receiver to report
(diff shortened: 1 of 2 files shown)


otel-collector/config.yaml
```diff
@@ -22,6 +22,20 @@ receivers:
...
+  # CVM host metrics — CPU, memory, disk, network, and load averages.
+  # root_path tells the receiver to read /proc and /sys from the host
+  # bind-mount rather than the container's own filesystem.
+  hostmetrics:
+    root_path: /host
+    collection_interval: 30s
+    scrapers:
+      cpu:
+      memory:
+      disk:
```
cl4wb0t, line 38: The `filesystem` and `disk` scrapers will pick up every mount and block device visible in the host namespace — overlay filesystems, tmpfs, `/proc`, `/sys`, cgroup mounts, loop devices, etc. In a Docker host that's easily 10+ virtual filesystems generating metrics every 30s, which inflates cardinality and GCP Cloud Monitoring cost for zero signal.

Consider adding exclude filters:

```yaml
      filesystem:
        exclude_mount_points:
          mount_points: ["/dev/*", "/proc/*", "/sys/*", "/run/*", "/var/lib/docker/*"]
          match_type: regexp
        exclude_fs_types:
          fs_types: ["tmpfs", "overlay", "squashfs", "proc", "sysfs", "devtmpfs", "cgroup2", "autofs"]
          match_type: strict
      disk:
        exclude:
          devices: ["loop.*", "fd.*", "sr.*"]
          match_type: regexp
```

This keeps real host disk/filesystem metrics and drops the container noise.

docker-compose.phala.yml
```diff
@@ -84,6 +84,11 @@ services:
       # Phala CVM secrets in the deployment manifest.
       GCP_SERVICE_ACCOUNT_JSON: ${GCP_SERVICE_ACCOUNT_JSON}
       GCP_PROJECT_ID: ${GCP_PROJECT_ID}
+    # Host filesystem + pid namespace for the hostmetrics receiver to report
```
cl4wb0t: Mounting the entire host root filesystem (`/:/host:ro`) gives the collector read access to everything on the host — `/etc/shadow`, private keys, any secrets on disk, etc. In a CVM handling cryptographic operations this is a wider surface than strictly necessary.

The hostmetrics receiver mainly needs `/proc` and `/sys`. The `filesystem` scraper also reads mount info but that's available via `/proc/mounts`. You could tighten this to:

```yaml
    volumes:
      - '/proc:/host/proc:ro'
      - '/sys:/host/sys:ro'
```

…and set `root_path: /host` (the receiver already prepends root_path to `/proc` and `/sys` lookups). The filesystem scraper would lose the ability to stat actual mount points for disk-usage numbers, but if you add the full-root mount back for that scraper specifically you can scope it more narrowly.

Not necessarily a blocker since this container already holds GCP service account credentials — just flagging it as a conscious security tradeoff worth documenting.

cl4wb0t (commented):
Overall this looks good — the hostmetrics receiver is the right call over the node-exporter sidecar. Two things worth considering before merge.

Also: the PR title still says "Node Exporter sidecar" but the implementation uses the built-in hostmetrics receiver — worth updating so the merge commit message matches what actually shipped.

Garandor (approved):

PR #153, opened 2026-03-20
feat: add dstack-ingress for custom domain TLS (CPL-118)

cl4wb0t:
## Summary

- Add `dstack-ingress` service to `docker-compose.phala.yml` behind `profiles: ["custom-domain"]` so current deployments are unaffected
- dstack-ingress terminates TLS inside the TEE with an attestation-bound cert covering both the per-node `NODE_DOMAIN` and shared `api.dev.litprotocol.com` (ALIAS_DOMAIN SAN)
- Uses Route 53 for DNS-01 ACME challenges (natively supported by dstack-ingress)
- Adds `cert-data` volume for Let's Encrypt cert persistence
- Documents new CVM secrets: `NODE_DOMAIN`, `CERTBOT_EMAIL`, `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`

**Blocked on:** [dstack-examples PR #86](https://github.com/Dstack-TEE/dstack-examples/pull/86) (ALIAS_DOMAIN support)

## Non-disruptive merge reasoning

This PR **can be safely merged** before R53 delegation and custom domain DNS are in place:

1. **Profile gating** — dstack-ingress is behind `profiles: ["custom-domain"]` and will not start unless explicitly activated with `docker compose --profile custom-domain up`. Standard `docker compose up` (used by current Phala deploys) ignores profiled services entirely.
2. **No changes to existing services** — lit-api-server, lit-actions, otel-collector, and lit-static are untouched.
3. **Volume addition is inert** — the new `cert-data` volume is only mounted by dstack-ingress, which doesn't start without the profile.
4. **Deploy workflow unaffected** — the `sed` image substitution step only touches `${DOCKER_IMAGE_*}` placeholders; `dstack-ingress` uses a literal image reference.

## Test plan

- [x] `docker compose -f docker-compose.phala.yml config` validates without errors (only expected env var warnings)
- [ ] Verify `docker compose up` (no profile) does **not** start dstack-ingress
- [ ] Verify `docker compose --profile custom-domain up` starts dstack-ingress (once PR #86 lands and R53 is configured)

Resolves CPL-118

### DIFF
--- a/docker-compose.phala.yml
+++ b/docker-compose.phala.yml
@@ -14,6 +14,12 @@
 # Required secrets (set as encrypted Phala CVM environment variables):
 #   GCP_SERVICE_ACCOUNT_JSON  - GCP service account key (raw JSON or base64-encoded)
 #   GCP_PROJECT_ID            - GCP project ID (e.g. "my-gcp-project")
+#
+# Additional secrets required for custom-domain profile (dstack-ingress):
+#   NODE_DOMAIN               - Per-node domain (e.g. "node1.api.dev.litprotocol.com")
+#   CERTBOT_EMAIL             - Email for Let's Encrypt certificate notifications
+#   AWS_ACCESS_KEY_ID         - AWS IAM key for Route 53 DNS-01 ACME challenges
+#   AWS_SECRET_ACCESS_KEY     - AWS IAM secret for Route 53 DNS-01 ACME challenges
 
 # RUST_LOG filter shared by lit-actions and lit-api-server.
 # App code stays at trace; per-module overrides suppress low-value internals:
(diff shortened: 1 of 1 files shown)


cl4wb0t (commented):
## Review

Compose profile approach is the right call — keeps dstack-ingress opt-in without affecting existing deployments. A few things to flag:

### 1. `image: dstacktee/dstack-ingress:latest` — mutable tag
The rest of the compose file uses `@sha256:` digest-pinned images (the header comment calls this out as DR-1.1/DR-1.2). `:latest` breaks that provenance model. Fine while blocked on PR #83, but should be pinned to a digest before production.

### 2. `ALIAS_DOMAIN` hardcoded to `api.dev.litprotocol.com`
This won't work for `chipotle-next` (different domain). Should this be `${ALIAS_DOMAIN}` passed via the deploy workflow, like `NODE_DOMAIN`?

### 3. Deploy workflow not updated
The deploy step does `sed` substitutions for image placeholders and `GCP_PROJECT_ID`. With the `custom-domain` profile, the workflow will eventually need to:
- Pass `--profile custom-domain` to activate the service
- Provide `NODE_DOMAIN`, `CERTBOT_EMAIL`, `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY` as secrets

Probably intentional (blocked on PR #83), but worth noting.

### 4. No health check on dstack-ingress
If dstack-ingress takes time to get its cert from Let's Encrypt, the NLB could start routing `:443` traffic before TLS is ready. A `healthcheck` on the container (e.g., checking nginx is listening or cert exists) might be worth adding.

### 5. `cert-data` volume 👍
Persists certs across container restarts so Let's Encrypt rate limits aren't hit on redeploys.

cl4wb0t (commented):
**Correction on point 4:** Ignore the health check comment — NLB health checks hit lit-api-server on :8000 directly (via `/health`), not through dstack-ingress on :443. dstack-ingress cert readiness doesn't affect health check availability.

cl4wb0t (commented):
**Strike the correction, point 4 stands.** The whole point of dstack-ingress is that it's the front door — NLB routes real traffic through :443 via dstack-ingress. Health checks should verify the same path real requests take. If dstack-ingress is down but lit-api-server is healthy on :8000, bypassing ingress gives a false positive.

Once the `custom-domain` profile is active, the NLB health check needs to go through :443, which means dstack-ingress must be ready (cert issued, nginx listening) before the node is marked healthy. A container-level `healthcheck` that verifies this would prevent routing traffic to nodes where cert issuance is still in progress.

cl4wb0t (commented):
**Final correction on point 4: it's invalid, disregard.** If the NLB health check goes through :443, the request traverses dstack-ingress before reaching `/health`. If dstack-ingress isn't up (no cert, nginx not listening), the TCP connection fails and the NLB marks the node unhealthy automatically. dstack-ingress being healthy is already an implicit prerequisite for the health probe to succeed — no separate container healthcheck needed.

PR #176, opened 2026-03-26
Remove hardcoded deployer keys from Makefile

Garandor:
## Summary
- Remove hardcoded testnet deployer private keys from the contract deployment Makefile
- `DEPLOYER_SECRET` must now be provided as an environment variable; Make errors clearly if unset
- The Anvil key (`0xac0974...`) is retained as a default since it is the well-known Hardhat/Foundry account #0

## Context
Prep for making the repo public (CPL-137). Two testnet deployer keys were hardcoded in the Makefile. They are not valuable but shouldn't be in source.

## Test plan
- [ ] `make deploy_sepolia` without env var → errors with "DEPLOYER_SECRET is not set"
- [ ] `DEPLOYER_SECRET=0x... make deploy_sepolia` → works as before
- [ ] `make deploy_anvil` → works without any env var (uses default Anvil key)

### DIFF
--- a/lit-api-server/blockchain/lit_node_express/Makefile
+++ b/lit-api-server/blockchain/lit_node_express/Makefile
@@ -4,9 +4,9 @@
 RUST_PROJECT := ../rust_generator_and_deployer
 GENERATOR_BIN := $(RUST_PROJECT)/target/debug/contract_generator
 DEPLOYER_BIN := $(RUST_PROJECT)/target/debug/contract_deployer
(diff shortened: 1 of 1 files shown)


Garandor (commented):
Reviewed — looks good, clean and minimal.

- `?= $(error ...)` is the right GNU Make pattern: deferred expansion means the error only fires when a target actually references the variable, so `make compile`/`make facade` still work without env vars.
- Keeping the well-known Anvil key as a default makes sense.

**Pre-existing note (non-blocking):** `--secret=$(...)` on the command line is visible in `ps` output. Worth a follow-up if the repo goes public.

**Reminder:** The keys in git history need to be rotated separately.

GTC6244 (approved):

PR #354, opened 2026-05-19
docs: design plan for lit-payments app

clawdbot-glitch003:
## Summary
- Adds `plans/lit-payments-app.md`: design doc for a separate ops-facing billing app (Discord-mod credit portal, LITKEY→Stripe credit via a Base payment gateway contract, deferred subscription billing).
- Architecture: new Rust crate in this monorepo, deployed on Railway outside the TEE, magic-link auth, Stripe restricted key, CoinGecko rate oracle, Alchemy WSS+poll listener.
- No code changes yet — this is a plan-only PR for review before scaffolding begins.

## Test plan
- [ ] Review the plan end-to-end; flag anything that should change before implementation starts.

PR #358, opened 2026-05-19
refactor: extract lit-billing-core from lit-api-server stripe.rs

clawdbot-glitch003:
## Summary
- New `lit-billing-core` crate owns the Stripe HTTP client, customer-by-wallet lookup, balance reads, reporting helpers, and pure formatters. `lit-payments` (coming separately) will depend on it so the customer-identity invariant — `metadata.wallet_address` keys the Stripe customer — is single-source-of-truth.
- `lit-api-server/src/stripe.rs` keeps its public API unchanged (caches, charge flows, PaymentIntent flow, `resolve_wallet_address`) and now delegates HTTP + customer/balance primitives to `lit-billing-core`.
- Pure refactor: zero behavior change. `cargo check`, `cargo test --lib` (211 passing), `cargo test --bin stripe_report` (14 passing), and `cargo clippy -D warnings` are all clean.

See `plans/lit-payments-app.md` (PR #354) for the bigger picture.

## Test plan
- [x] `cargo check` clean in both crates
- [x] `cargo test --lib` — 211 lit-api-server tests pass
- [x] `cargo test --bin stripe_report` — 14 tests pass
- [x] `cargo test` in `lit-billing-core` — 15 tests pass (moved from stripe.rs)
- [x] `cargo clippy --all-targets -- -D warnings` clean in both crates
- [x] `cargo fmt --check` clean


GTC6244 (approved):
looks good - cache stays it lit-api-server, which was my #1 concern. ;-)

PR #366, opened 2026-05-20
fix(CPL-326): tolerate git-describe abbrev drift in wait-for-api-restart

GTC6244:
## Summary
- `commit_version` comes from `git_version!()` (`git describe --always`). With `core.abbrev=auto`, the abbreviated SHA can be one character longer or shorter depending on the local repo's object count, so the builder container and the workflow runner sometimes disagree (e.g. `g615bf66` vs `g615bf664`) and the polling loop in `wait-for-api-restart.yml` times out waiting on an exact string match.
- Compare with prefix match in either direction: the shorter SHA is always a prefix of the longer one for the same commit, so an off-by-one abbrev now passes while genuine mismatches still fail.

## Test plan
- [ ] Trigger a deploy that calls `wait-for-api-restart.yml` and confirm the version match succeeds when builder/runner abbrev lengths differ.
- [ ] Verify the workflow still times out correctly when the API never returns the expected SHA.

### DIFF
--- a/.github/workflows/wait-for-api-restart.yml
+++ b/.github/workflows/wait-for-api-restart.yml
@@ -118,6 +118,13 @@ jobs:
           # The API's commit_version is produced by git_version!() which runs
           # git describe --always at compile time. Reproduce the same output
           # here so we can compare directly.
+          #
+          # Note: git describe abbreviates the SHA using core.abbrev=auto
+          # (default since Git 2.11), which picks length based on the local
+          # repo's object count. The builder container and this runner can
+          # differ by a character (e.g. g615bf66 vs g615bf664), so we accept
+          # any pair where one side is a prefix of the other — the shorter
+          # SHA is always a prefix of the longer one.
           EXPECTED_VERSION=$(git describe --always "${{ inputs.expected_sha }}")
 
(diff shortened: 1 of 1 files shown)


GTC6244:
CI status (at time of triage):
- Socket Security checks: passing.
- Rust CI, k6-client-check, contract-bindings, Phala simulator: still in progress — no source code in those areas changed, so I'd expect them to pass.
- Copilot review: informational summary only, no actionable findings. No changes required from the review.

`mergeStateStatus` is `BLOCKED` pending the in-progress checks and human approval; nothing to fix on my end right now.

GTC6244:
Re-checked CI:
- `lit-actions` Rust CI failed on a single integration test, `pool_warm_hit`, with `expected pool hit (hits before=0, after=0, target=10)`. That's a timing-dependent assertion on the warm Deno pool from CPL-264 (commit 462a9d21), not anything this PR touches — the diff here is workflow-YAML-only. A concurrent PR (`feature/cpl-260-analyze-lit-core`) hit the same failure at the same time and went green on retry, which matches the flake pattern.
- All other checks passed (`contract-bindings`, `k6-client-check`, `dstack simulator attestation`, the rest of Rust CI, Socket Security).
- I've re-run just the failed Rust CI job.

If `pool_warm_hit` keeps flaking, worth filing a separate ticket against the warm-pool test rather than blocking this PR.

PR #383, opened 2026-05-25
feat(e2e): Playwright suite for the dashboard (NODE-4968)

GTC6244:
## Summary

- Adds a Playwright suite under `e2e/` covering the dashboard at `lit-static/dapps/dashboard` in all three auth modes — API key, ChainSecured/EOA (MetaMask via Synpress), and ChainSecured/WalletConnect (headless `@reown/walletkit`) — and exercises the same surface as the k6 suite (chain config, account + usage-key creation, hello-world / encrypt-decrypt / ecdsa-sign Lit Actions, IPFS CID lookup).
- Wires the suite to the existing `local_test.sh` stack via a Makefile, and adds a Playwright `globalSetup` that registers lit-api-server's dstack-derived api_payer wallet on the AccountConfig diamond + funds it, closing a setup gap that otherwise blocks `/new_account` against a fresh local deploy.
- Two small enabling changes outside `e2e/`: `lit-static/wallet_connect.js` re-emits the WalletConnect SDK's `display_uri` event as a DOM `CustomEvent` so tests can pair a headless wallet without scraping the QR modal (dashboard ignores it); `local_test.sh` makes the Jaeger/Docker step optional (telemetry only) so the stack boots cleanly when Docker isn't running.

## Test plan

- [ ] `make -C e2e install`
- [ ] `make -C e2e up` (boots `local_test.sh` in the background)
- [ ] `make -C e2e test-api` — pure HTTP/UI, no wallets
- [ ] `WC_PROJECT_ID=… make -C e2e test-wc`
- [ ] `make -C e2e test-eoa` — requires a working Synpress wallet cache (`pnpm run build:wallet-cache`)
- [ ] `make -C e2e typecheck`


GTC6244:
Thanks for the review — addressed all 8 in `cf8b6a90`:

1. **`flows/onboarding.spec.ts` pulling in `wcWallet` for the API path** — split into `flows/api-onboarding.spec.ts` and `flows/wc-onboarding.spec.ts`, each picked up by its own project via `testMatch`. Dropped the now-unused `walletKind` fixture option from both `test.ts` and `test-with-mm.ts`.
2. **Missing visibility check before `#tx-preview-confirm` in the WC flow** — added `expect(previewConfirm).toBeVisible({ timeout: 45_000 })` to match the dedicated WC specs.
3. **`basic.setup.ts` comment claiming account #1** — comment updated to match Synpress's actual default (account #0) and to point at `metamask.switchAccount(...)` for tests that need separation.
4. **`runLitAction` waiting on the global 15s expect timeout** — now passes an explicit 60s timeout, comfortably under the 90s test timeout.
5. **`getLitActionIpfsCid` same issue** — now passes 30s explicitly.
6. **`display_uri` re-emission exposing the pairing URI on any origin** — gated on `window.location.hostname` ∈ `{localhost, 127.0.0.1, ::1}`. Production origins never emit it.
7. **README referencing a `flows-eoa` project that doesn't exist** — table rewritten to show the actual `flows-api` / `flows-wc` split and explain why EOA lives under `tests/eoa/`.
8. **README listing Docker as required** — moved to "optional, telemetry only" and called out that nothing in the suite depends on Jaeger.

Verified `pnpm exec tsc --noEmit` and `pnpm exec playwright test --list` (14 tests across 5 projects).

GTC6244:
Re-checked the PR — no new comments since the response above; everything from the prior review is addressed in cf8b6a90.

PR #388, opened 2026-05-26
chore(lit-actions): upgrade Deno runtime stack

clawdbot-glitch003:
## Summary
- Upgrade the `lit-actions` Deno crate stack to Deno v2.8.0-era versions (`deno_core`, `deno_runtime`, `deno_lib`, `deno_resolver`, `deno_error`, `sys_traits`)
- Update SWC pins and Rust code for upstream API drift in ops, snapshots, module loading, workers, error formatting, and SWC WTF-8 strings
- Preserve runtime behavior for Lit Actions by wiring residual lazy extension sources from snapshot creation into worker startup
- Fix fetch/TLS under rustls 0.23 by explicitly installing the aws-lc-rs crypto provider and using the no-provider reqwest feature
- Adjust integration warmup timing for the slower upgraded Deno boot path under concurrent tests

## Validation
- [x] `git diff --check`
- [x] `cargo fmt --all`
- [x] `cargo test -p lit-actions-server --lib -j 2` — 86 passed
- [x] `cargo test -p lit-actions-tests --test integration -j 2` — 26 passed, 2 ignored
- [x] `cargo test --workspace -j 2`

## Notes
- Local root disk is tight after rebuilding Rust targets: `/` was 98% used with ~3.5G free during validation.
- The Deno stack update removes the previous Lit fork patch for `deno_core` and the macOS libffi patch because the upgraded dependency graph no longer needs them.


lit-actions/server/runtime.rs
```diff
@@ -35,6 +36,24 @@ use tokio::sync::{mpsc, oneshot};
 use tonic::Status;
 use tracing::{debug, error, info_span, instrument};
 
+include!(concat!(env!("OUT_DIR"), "/runtime_lazy_sources.rs"));
+
+fn lit_actions_ops_extension() -> deno_core::Extension {
+    let mut extension = lit_actions_ext::lit_actions::init();
+
+    // The startup snapshot is built with the full Lit Actions extension. At
```
GTC6244, line 44: I'm assuming this skips all of the manual ESM work that I had Claude put together - but will need to load test, and/or profile to be sure.

lit-actions/tests/it.rs
```diff
@@ -894,12 +894,14 @@ fn pool_test_setup() -> (TestClient, Arc<PoolHealth>, usize) {
 
 /// Coarse warmup wait. There's no "ready workers" counter today, so we
 /// just sleep long enough for snapshot-bootstrapped workers to land in
-/// the ready channel. With pool=10, 600ms is plenty under load.
+/// the ready channel. The newer Deno runtime has more lazy extension state to
```
GTC6244: While this is in the test path, the comment makes me believe that timing may be changing in the processing.    On the one hand, reducing the TTL of the actual runtime is good, but lazy-loading after the fact may ( or may not ) make the entire process a bit slower.   Suspect this is a nanosecond thing - but for clients like Relay, this may make a difference. 

To be more specific, we're caching deno environments entirely, so I'd prefer to do as much hydration prior to deno saying it's ready to process ! 

Again, profiling is the answer here.

GTC6244 (commented):
@glitch003 - looks pretty good and I'd give it a conditional pass, based on the fact that it works !    

My notes inline are really about running profiling tests, which I'll do today.

GTC6244:
@copilot resolve the merge conflicts in this pull request

GTC6244:
Very oddly, the lit-actions server with the new Deno runtime doesn't want to execute on my local machine due to an issue /w tokio.    (Compilation works, execution does not ).     Likely a mac-only problem, as tests here pass.

GTC6244:
From Claude: 

Root cause. Your Deno upgrade (deno_core 0.338 → 0.401) introduced a new libuv/Node TTY compatibility layer. During MainWorker::bootstrap_from_options, deno_core registers process stdin/stdout/stderr by calling tokio::io::unix::AsyncFd::new (uv_compat/tty.rs:1343), which requires a Tokio runtime with the IO driver on the current thread. Bootstrap worked fine in 0.338 without one.

GTC6244:
Ok - further reading: 

The bug was latent, not Mac-only. CI and prod just never exercise the TTY path because their stdio is always redirected. Anyone running it locally in a terminal — on any OS — would hit it.

The fix is still worth keeping. It makes warm-time bootstrap correct regardless of how stdio is wired (entering the runtime before bootstrap costs nothing), and it removes the local-dev footgun. It’s the architecturally right thing anyway — bootstrap and the request event loop now share one reactor instead of bootstrapping bare and running on a throwaway runtime.

This makes sense, and will commit.

GTC6244 (approved):

PR #431, opened 2026-06-02
fix(lit-actions): stop cross-execution js_params/bytecode leak via the eval-context code cache

clawdbot-glitch003:
## Summary

A Lit Action could run with **another request's `js_params`** — and, more
broadly, with **another action's compiled bytecode**. The eval-context V8 code
cache keys on a hash that collides for equal-length sources, so two unrelated
executions could be served each other's cached compilation. It caused
intermittent wrong-input execution (a cross-request data mix) and, secondarily,
fatal `V8_Fatal: stack_overflow()` aborts of the worker.

## Root cause

User code runs via `__litEvalCached(source, specifier)` → `op_eval_context` →
the `SharedV8CodeCache`, whose key is `(specifier, kind, source_hash)`.

Two things made that key collide across unrelated executions:

1. **`source_hash` degenerates to a length-based hash for large sources.** Deno's
   `op_eval_context` computes the *same* `source_hash` for different sources of
   equal length. V8's own `ConsumeCodeCache` validation is itself length-based,
   so it does not reject a mismatched entry either.
2. **The specifier was a constant** (`file:///user_provided_script.js`), so every
   action shared one keyspace.

On top of that, **per-request `js_params` were baked into the executed source**
by `CachedActionCode::to_executable_code`:

```js
{bundled action code}
;(async () => {
  const params = {js_func_params};   // <-- per-request params baked INTO the source
  const data = await main(params);
  ...
})();
```

So there were two distinct ways one execution could receive another's cached
compilation:

- **Params crossing:** the *same* action invoked with *different* params of equal
  serialized length → equal-length sources → `source_hash` collision → a cache
  **HIT** returns the first request's bytecode, with its baked-in params → the
  action runs **its own code but the first request's `js_params`.**
- **Action crossing:** *different* actions whose bundled sources happen to be
  equal length → `source_hash` collision under the shared constant specifier →
  one action is served the **other action's compiled bytecode.**

### Evidence (reproduced on a full local stack: anvil + dstack-sim + lit_actions + lit-api-server)

Instrumented client → lit-api-server boundary → action, matched by `sessionId`:
- client sends and lit-api-server forwards `sid=56634041 round=2` correctly,
- but the action executes with `sid=dc6e621d` — a *different* request's params.

Logging the cache directly: two requests with **different source content**
(`myhash` 86e200e7 vs 48f21109) but **equal length** (38164 bytes) produced the
**identical** Deno `source_hash=4decc2ac…` and both HIT.

Reproduced reliably with `examples/mpc-signing`'s 2-of-3 keygen (it relays large,
similar-sized session blobs round to round, so collisions are frequent). It
surfaced as `Invalid commitment hash` / `Missing message` (the MPC protocol
legitimately rejecting the wrong session) plus occasional V8 stack-overflow
aborts.

## Fix

Two complementary changes; both are required to close the bug.

**1. Stop baking params into the code-cached source.** Inject `js_params` as a
single global, `globalThis.__litJsParams`, via `execute_script` (which bypasses
the eval-context cache), and have the per-execution wrapper read it:
- `inject_params_globals` now sets `globalThis.__litJsParams = <params>` (was a
  no-op `Object.assign`, always called with `None`).
- `execute_with_worker` calls it with the actual `js_params` at the existing
  reserved hook.
- `to_executable_code` reads `const params = globalThis.__litJsParams` instead of
  inlining the params.

This makes the source compiled through `op_eval_context` the **stable, param-free
bundled action**, so the same action no longer crosses params on a `source_hash`
collision.

**2. Key the code cache on a content-derived specifier.** The eval stub now passes
`file:///user_provided_script_<ipfs_id>.js` (the action's IPFS CID — already the
canonical content id used for the action-code cache) instead of the constant
specifier:
- `runtime.rs`: content-derived specifier passed to `__litEvalCached`.
- `lib.rs`: expose `get_lit_action_ipfs_id` (the canonical content id).
- `v8_code_cache.rs`: `specifier_is_part_of_key` unit test (same hash + kind,
  different specifier → independent entries).
- `it.rs`: stack-trace assertions compute the specifier dynamically via
  `get_lit_action_ipfs_id`; bump `99_patches.js` call-site line for the new
  comment.

Each action now occupies its own keyspace: **same action → same specifier →
still reuses bytecode** (the CPL-264 optimization is preserved); **different
action → different specifier → a clean miss**, so V8 is never offered a
mismatched cache entry to accept.

### Why both

The two vectors are independent, so neither change alone is sufficient:
- Change 1 alone (param-free source, constant specifier) still lets two
  *different* actions of equal length collide and cross **bytecode**.
- Change 2 alone (content-derived specifier, params still baked in) still lets the
  *same* action with equal-length params collide and cross **params**.

Together there is no remaining path for one execution to receive another's
bytecode or params.

Individual param keys are not promoted to globals (passed to `main(params)`
exactly as before; `globalThis.Hello`-style access stays `undefined`). The whole
object is bound to a single internal global, `globalThis.__litJsParams`, for the
duration of the execution. Within one (single-tenant, fresh-per-request) isolate
that object is therefore readable by the action's own bundled code/imports — an
intra-execution defense-in-depth gap, not a cross-request path. Fully hiding it
would require emitting a statement before the action body in the per-execution
wrapper, which demotes a user action's own top-level `"use strict";` and silently
switches their code to sloppy mode; the alternatives (forcing whole-script strict,
or wrapping the body in a function) likewise change sloppy-mode and top-level
`var`/`this` semantics for existing actions. Since this is a same-trust-domain
hardening nicety, it is an accepted limitation rather than worth a semantics
regression.

### Behavior note

The specifier appears in user-facing error stack traces, so traces now carry the
action's IPFS id in the filename (`file:///user_provided_script_<id>.js`). A trace
only ever shows the executing action's own id, which its caller already knows.

## Verification

- 2-of-3 MPC keygen against the full local stack: **15/15, zero retries, no crash**
  (was ~25–50% per-attempt failure + intermittent worker aborts).
- `lit-actions` integration suite green (incl. `js_params` and the updated
  stack-trace assertions), plus the new `v8_code_cache` unit test.
- All PR CI checks green.

## Severity

Not MPC-specific: **any** two Lit Action executions that collide on the
length-based `source_hash` can receive each other's `js_params` or compiled
bytecode — worth treating as a potential cross-request/cross-tenant
data-exposure bug, not just flakiness.

### DIFF
--- a/lit-actions/ext/js/99_patches.js
+++ b/lit-actions/ext/js/99_patches.js
@@ -39,6 +39,10 @@ globalThis.LitTest = { op_panic };
 // parses per execution but its body is one string literal, so V8 only pays
 // for source-string scanning, not for compiling the bundled action body.
 //
+// The caller passes a content-derived `specifier` (the action's IPFS id) so
+// distinct actions occupy distinct code-cache keys; see the call site in
+// runtime.rs for why a shared specifier would let equal-length actions collide.
+//
 // The helper deletes itself from globalThis as its first action so user code
 // (which runs inside op_eval_context below) cannot reach it; this preserves
 // the `--disallow-code-generation-from-strings` posture by not handing actions
--- a/lit-actions/server/lib.rs
+++ b/lit-actions/server/lib.rs
@@ -7,7 +7,7 @@ pub mod worker_pool;
 
 pub mod server;
 
-pub use runtime::init_v8;
+pub use runtime::{get_lit_action_ipfs_id, init_v8};
 pub use server::*;
 
 // Re-exports
--- a/lit-actions/server/lib.rs
+++ b/lit-actions/server/lib.rs
@@ -2,7 +2,7 @@ mod bundler;
 pub mod cdn_module_loader;
 mod import_rewriter;
 mod runtime;
-mod v8_code_cache;
+pub mod v8_code_cache;
 pub mod worker_pool;
 
 pub mod server;
--- a/lit-actions/server/runtime.rs
+++ b/lit-actions/server/runtime.rs
@@ -261,7 +261,12 @@ fn deno_isolate_init() -> Option<&'static [u8]> {
     Some(RUNTIME_SNAPSHOT)
 }
 
-fn get_lit_action_ipfs_id(code: &str) -> String {
+/// Compute the IPFS CID for a Lit Action's source. This is the action's
+/// canonical content identity: it keys the action-code cache and, via the
+/// op_eval_context specifier, the V8 code cache (so distinct actions can't
+/// collide on V8's length-based source hash). Exposed for tests that assert on
+/// the specifier (it appears in user-facing stack traces).
+pub fn get_lit_action_ipfs_id(code: &str) -> String {
     let ipfs_hasher = IpfsHasher::default();
     ipfs_hasher.compute(code.as_bytes())
 }
@@ -836,10 +841,22 @@ async fn execute_with_worker_inner(
     // execution, but its body is one string literal, so V8 only pays for
     // source-string scanning rather than compiling the bundled action body
     // (CPL-264).
+    //
+    // The specifier is content-derived (the action's IPFS id) rather than a
+    // constant. Deno's code-cache key is `(specifier, kind, source_hash)`, and
+    // `source_hash` degenerates to V8's length-based string identity hash for
+    // large sources — so two *different* actions of equal length would share a
+    // `source_hash` and collide under a shared specifier, handing one action
+    // the other's compiled bytecode. Keying the specifier on the action id puts
+    // each action in its own keyspace: same action -> same specifier -> still
+    // reuses bytecode (CPL-264); different action -> different specifier -> a
+    // clean miss, so V8 is never offered a mismatched cache entry to accept.
     let user_code_literal =
         serde_json::to_string(&user_code).context("Could not serialize user code for eval stub")?;
-    let stub =
-        format!("__litEvalCached({user_code_literal}, \"file:///user_provided_script.js\");");
+    let specifier = format!("file:///user_provided_script_{action_ipfs_id}.js");
+    let specifier_literal = serde_json::to_string(&specifier)
+        .context("Could not serialize eval specifier for eval stub")?;
+    let stub = format!("__litEvalCached({user_code_literal}, {specifier_literal});");
 
     if let Err(e) = worker
         .js_runtime
--- a/lit-actions/server/runtime.rs
+++ b/lit-actions/server/runtime.rs
@@ -187,13 +187,18 @@ impl CachedActionCode {
                 .sum::<usize>()
     }
 
-    fn to_executable_code(&self, js_func_params: &str) -> String {
+    // NB: js_params are NOT baked in here — they are injected separately as the
+    // `globalThis.__litJsParams` global (see `inject_params_globals`). Keeping
+    // them out makes this source stable per bundled action, so the
+    // eval-context code cache keys/compiles correctly across requests instead
+    // of crossing one request's params into another's execution.
+    fn to_executable_code(&self) -> String {
         format!(
             "
         {code}
         ;
         (async () => {{
-        const params = {js_func_params} ;
+        const params = globalThis.__litJsParams;
         const data = await main(params);
         if (typeof data !== \"undefined\") {{
           LitActions.setResponse( {{ response: data }} );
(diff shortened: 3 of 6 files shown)


clawdbot-glitch003:
Ran an independent review + a Codex (`gpt-5.1-codex`) pass on the latest state. Core fix looks correct and safe — both crossing vectors (params via the param-free cached source, bytecode via the content-derived specifier) are closed, and they compose. Codex's gate passed with no critical findings. It raised two **P2** advisories, both of which I verified and fixed in 0d2c2015:

**1. Preserve `null` for omitted `js_params`.** `inject_params_globals` returned early when `js_params` was absent, so `globalThis.__litJsParams` stayed `undefined` and `main` received `undefined`. The pre-change path serialized `Value::Null` (`main(null)`), so actions that branch on `params === null` for the code-only / no-params case would have silently regressed. Now we inject `null` explicitly when params are absent — params stay out of the cached source, and the no-params semantics match the old behavior.

**2. Inject params *after* the resource guards are armed.** The injection ran via `execute_script` at the top of `execute_with_worker_inner`, before `start_controller_thread` and `add_near_heap_limit_callback` were registered. Materializing a large `js_params` object into V8 (especially under a small custom `memory_limit` or short timeout) could therefore escape our timeout/OOM handling and hit V8's hard heap limit, hard-aborting the worker instead of returning the normal `resource_exhausted`/timeout error. Since this whole PR is partly about turning a hard `V8_Fatal` abort into graceful handling, that ordering mattered. Moved the call below both guards — it only needs to run before the user-code stub, and there's no dependency on the controller thread.

Verification:
- `cargo check -p lit-actions-server` clean; `v8_code_cache` unit tests (incl. `specifier_is_part_of_key`) green.
- `js_params` and `lit_eval_cached_is_hidden_from_user_code` integration tests pass against the change.
- (Note: `reference_error` fails locally for me with a byte-identical stack-trace diff — a `pretty_assertions` whitespace/color artifact in my local env; it fails the same way on the unmodified branch tip and is green in CI, so it's unrelated to this change.)

Also refreshed the PR description/title to cover the content-derived-specifier commit (it previously only described the `js_params` global-injection part).

clawdbot-glitch003:
Fixed Codex's P2 finding A (params exposed as a global) in 12084097.

**The issue:** `js_params` were injected as `globalThis.__litJsParams` and left on the global for the whole execution. The bundled action's top-level code and any imported CDN dependency could therefore read — or overwrite — the full params object before `main(params)` ran, including params the author never intended to hand a given import. Pre-cache, params were a closure-local literal invisible to that code, so this was a real regression in the "params are not globals" posture (and it contradicted the PR description's claim).

**The fix:** the per-execution wrapper now captures the injection global into a script-local binding and `delete`s `globalThis.__litJsParams` as its **first statement**, before the bundled action code (and its imports) run:

```js
const __litActionParams = globalThis.__litJsParams; delete globalThis.__litJsParams;
{bundled action code}
;
(async () => {
  const params = __litActionParams;
  const data = await main(params);
  ...
})();
```

Params reach only `main(params)`. The capture replaces the wrapper's previously-empty first line, so the bundled action still starts on line 2 — the `reference_error` / `throw_error` stack-trace assertions are unchanged (verified the actual frames are still `2:33 / 6:28 / 10:11` etc.).

**Test:** extended the `js_params` test to assert `typeof globalThis.__litJsParams === "undefined"` inside user code; passes.

Verification: `cargo build -p lit-actions-server` clean; `js_params` and `lit_eval_cached_is_hidden_from_user_code` pass.

**On finding B** (timeout/OOM during params injection surfaces as a generic error instead of `resource_exhausted`/`deadline_exceeded`): leaving as-is for now. The earlier guard-ordering change already removed the hard `V8_Fatal` abort — this is a narrow edge (params large enough to OOM/timeout during their own materialization, i.e. a tiny custom `memory_limit` or pathological params) whose only symptom is a less-specific error message. Happy to wire the injection through the controller's classification path in a follow-up if you'd like it fully closed.

clawdbot-glitch003:
Correction / heads-up on finding A.

I reverted the two commits that attempted to hide `__litJsParams` from user code (`12084097`, `068a8798`) — the branch is back at `0d2c2015`. A deeper codex pass plus the integration suite showed those attempts were net-negative:

- To stop bundled/imported code reading the injected params, the params (or a closure over them) have to be neutralized **before** `{code}` runs — which means emitting a statement ahead of `{code}` in the per-execution wrapper.
- But emitting anything before `{code}` demotes a user action's own top-level `"use strict";` from a directive to a no-op expression, silently switching their code to **sloppy mode**. That regressed `lit_namespace_protection` (the test relies on strict-mode `TypeError`s when user code tries to delete/reassign the frozen `Lit` namespace).

So "fully hide params from same-execution bundled code" conflicts with "preserve the user's strict/sloppy-mode choice." Closing it the other ways (forcing whole-script strict, or wrapping `{code}` in a function) changes sloppy-mode and top-level `var`/`this` global-scoping semantics for existing actions.

Net: the **cross-request / cross-tenant** leak (the actual bug this PR fixes) remains fully fixed at `0d2c2015` via the content-derived specifier + param-free cached source. Finding A is an **intra-execution defense-in-depth** gap (an action's own untrusted import could read that action's own params via `globalThis.__litJsParams`) — same trust domain, fresh isolate per request. Fully closing it needs a design decision on the strict/sloppy tradeoff, so I'm leaving it out of this PR rather than shipping a semantics regression.

PR #437, opened 2026-06-02
ci(deploy): zero-downtime ping-pong cutover for staging

clawdbot-glitch003:
## What this does

Make staging deploys **zero-downtime** by ping-ponging between two long-lived instances of the **same Phala app** (`chipotle-next`, one `app_id` → one KMS identity / shared keys) on two nodes (prod2 + prod4). Each deploy targets the **cold** instance; the freshly-deployed CVM's `dstack-ingress` **auto-claims** the custom domain on boot; the old instance is stopped once the new one is verified serving. If post-cutover checks fail, it auto-rolls-back.

`main` is now the deploy branch and targets `chipotle-next` (the retired `next` branch's role moved to `main`; the dead-end `chipotle-dev` path is removed).

## The cutover flow

1. **`determine-target`** — derives the LIVE instance from DNS (the public domain's CNAME → live node's gateway, matched to each candidate via `phala cvms get`). Targets the **other** (cold) instance. Refuses to deploy onto the live box; errors instead of guessing if live can't be determined. There is **no stored live-pointer** — state is read from reality every run, so it can't drift, can't be corrupted by a cancelled run, and can never target the serving box.
2. **`deploy`** to the cold instance. Its ingress auto-claims the domain once `lit-api-server` passes `/health` (compose `service_healthy` gate).
3. **verify** on the cold box's **own default Phala URL** (not the shared domain): version match, attestation, openapi, k6 smoke + correctness.
4. **On success:** `confirm-cutover` (public domain serves the new SHA) → `stop-old-cvm`, which re-resolves the domain and **refuses to stop the old box if the domain still points at it** (no stranding).
5. **On failure:** `rollback-on-failure` — flips the public CNAME back to the still-running old instance and stops the bad new box. Both instances share one app_id, so only the CNAME moves (the `_dstack-app-address` TXT is identical); single Route 53 write via the existing `CERTBOT_AWS_*` role. Runs only when the new box came up healthy (if it never came up, it never claimed and the old box still serves).

Both instances stay as warm `stop`ped rollback targets (we `stop`, never `delete`), so deploys are **in-place upgrades** — the `cert-data` volume persists, so no new Let's Encrypt issuance per deploy (avoids the 5-cert/week/hostname limit).

## Files changed

- **`.github/workflows/deploy-staging.yml`** — `main`→`chipotle-next`; DNS-derived cold-instance selection; verify against the cold box's direct URL; `confirm-cutover` + `stop-old-cvm` (with move-off guard) + `rollback-on-failure`; `concurrency.cancel-in-progress: false` (don't cancel a multi-step cutover); no-op pushes skip the verify chain.
- **`docker-compose.phala.yml`** — `lit-api-server` `/health` healthcheck + `dstack-ingress depends_on: service_healthy` (domain only flips once the API is ready); `SET_CAA=false` (two in-TEE ACME accounts on one domain must not lock each other out of cert issuance).
- **`Dockerfile.lit-api-server`** — add `curl` for the healthcheck.
- **`.github/workflows/wait-for-api-restart.yml`** — expose a `base_url` output (so attestation targets the cold box).
- **`plans/zero-downtime-pingpong-deploy.md`** — design doc + Phase 0 findings.

## Required config

- Repo vars `NEXT_CVM_A` / `NEXT_CVM_B` — the two `chipotle-next` instance names (set: `chipotle-next-rep-sa6xj` @prod2, `chipotle-next-rep-sa6xj-rep-w4hdl` @prod4). Live instance is derived from DNS; no live-pointer var.
- Both nodes' Phala devices allowlisted on the app's AppAuth contract (`phala cvms device-allowlist`) so the shared app_id derives keys on either node.
- Existing `CERTBOT_AWS_*` secrets (already used for DNS-01) — reused by the rollback Route 53 write.

## Validated (Phase 0, dev zone) — see plan doc

Cert coexistence (two ACME accounts, one domain), in-CVM cert issuance via assumed role, custom-domain routing on `dstack-base` gateways, the auto-claim cutover + Route 53 flip/rollback mechanics, and that the gateway serves a domain from a single instance (not load-balanced). The single-`app_id` model is proven by the manual prod cutover (prod4→prod2) — it works because both nodes' devices are on-chain allowlisted (an earlier `phala cvms replicate` test failed only because it skipped `addDevice`).

## Hardened via adversarial (codex) review

Findings addressed: stateless live-detection (no var drift / missing-token / cancel corruption), `stop-old-cvm` move-off guard, `cancel-in-progress: false`, `SET_CAA=false`, no-op guard, and post-cutover auto-rollback. Full thread in the PR comments.

## Not done / decisions

- **`/health` depth (codex #8):** `/health` treats billing keys as informational, so a box with missing deps can still pass the boot-time claim gate. Deepening it (touches `health.rs`) makes the gate more meaningful and the rollback fire less often — left as a separate app-code change.
- **Architecture fork (codex #1):** the cutover happens at container boot (on `/health`), before the deeper CI checks — so those are post-hoc, covered by the auto-rollback (Option A). Option B (CI-owned DNS, claim only after full verification) would eliminate the boot→verify exposure window but needs a dstack-ingress cert-only patch. Shipping A; revisit B for prod if the window is unacceptable.

## ⚠️ Not yet CI-tested

Reviewed by eye + YAML-valid + adversarially reviewed, but not executed. The first real `main` deploy is the test. Key thing to watch: does `phala deploy --cvm-id <stopped cold CVM>` boot the stopped instance on the new compose? Everything downstream assumes it does; if not, the deploy fails safely (old box keeps serving) and we add explicit start handling.


clawdbot-glitch003:
**Update — branch→app mapping resolved + ping-pong armed.**

- `main` now deploys to **chipotle-next** (test.chipotle.litprotocol.com) with the ping-pong cutover; the retired `next` push trigger and the dead-end `chipotle-dev` path are removed. (pushed)
- Set the repo vars (inert until this merges):
  - `NEXT_CVM_A = chipotle-next-rep-sa6xj` (prod2)
  - `NEXT_CVM_B = chipotle-next-rep-sa6xj-rep-w4hdl` (prod4)
  - `NEXT_LIVE_CVM = chipotle-next-rep-sa6xj-rep-w4hdl` (prod4 is live now)

**Merging this PR is the test:** the merge is a push to `main` and touches `deploy-staging.yml` (a deploy-relevant path), so it will trigger a deploy → cold = **prod2 (sa6xj)** → verify on prod2's direct URL → confirm `test.chipotle` serves the new SHA → stop prod4 → set `NEXT_LIVE_CVM = sa6xj`.

Watch the **deploy step** on the first run: prod2 (sa6xj) is currently *stopped*, so this confirms whether `phala deploy --cvm-id <stopped>` boots it on the new compose. If it doesn't, the deploy fails safely (prod4 keeps serving, no downtime) and we add explicit start handling.

Still missing for the auto live-pointer update: a `PINGPONG_VARS_TOKEN` PAT (Variables: write). Without it, `record-live-cvm` warns + skips and `NEXT_LIVE_CVM` must be flipped by hand after each deploy.

clawdbot-glitch003:
**Adversarial (codex) review → hardening pushed.** Fixes in `f7df679`:

- **Stateless live detection (kills the var-drift class: codex #3/#4/#5).** `determine-target` now derives the live instance from the public domain's CNAME (→ live node gateway, matched to each candidate via `phala cvms get`) instead of a stored `NEXT_LIVE_CVM`. It refuses to deploy onto the live box, and errors (doesn't guess) when live can't be determined. Removed `record-live-cvm` + the `PINGPONG_VARS_TOKEN` requirement, and deleted the `NEXT_LIVE_CVM` repo var.
- **stop-old-cvm guard (codex #2).** Re-resolves the domain and refuses to stop the old instance if the domain still points at its gateway — prevents stranding the domain on a same-commit redeploy where both boxes share a SHA.
- **concurrency `cancel-in-progress: false` (codex #4).** Don't cancel a multi-step cutover midway.
- **No-op guard (codex #10).** `wait-for-api-available` only runs when a deploy happened, so no-op pushes don't poll the stopped cold box and time out.
- **`SET_CAA=false` (codex #7).** Two in-TEE ACME accounts on one domain no longer risk locking each other out of cert issuance.

### ⚠️ Decision needed — (d) what gates traffic (codex #1)

The deepest finding: the cutover happens at **container boot** (when `/health` passes, via the compose `service_healthy` gate), which is **before** CI runs attestation / openapi / k6 against the cold box. So those deeper checks are **post-hoc** — a `/health`-green but functionally-broken build can take public traffic before k6/attestation prove it. This is inherent to the auto-claim approach; it's a real fork, not a one-liner:

- **Option A — keep auto-claim, add post-cutover auto-rollback.** Simple deploy (what we have). If post-cutover k6/attestation fail, automatically flip back (the old box is still a warm `stop`ped rollback). Window of exposure = boot→k6 (~minutes) where a broken-but-/health-green build serves. Mitigate by making `/health` deeper (codex #8: it currently treats billing keys as informational).
- **Option B — CI-owned DNS, claim only after full verification.** Ingress runs cert-only (doesn't auto-write routing records); CI flips the CNAME/TXT itself *after* attestation + k6 pass on the cold box. Zero exposure to unverified traffic, but needs the dstack-ingress cert-only patch (the approach the plan's §4 called Option 1) — more work.

Current code implements **A** (minus the auto-rollback). My recommendation: ship **A** for the staging dogfood to prove the mechanics, decide on auto-rollback + a deeper `/health` before prod, and only build **B** if the boot→verify exposure window is unacceptable for prod. Flagging for your call.

clawdbot-glitch003:
**Decision (d) → Option A implemented: post-cutover auto-rollback.** (`3f0d1d5`)

New `rollback-on-failure` job: if the deeper checks (attestation / openapi / k6) fail *after* the cold box came up and auto-claimed the domain, it flips the public domain's CNAME back to the still-running old instance and stops the bad new box. Since both instances share one app_id, only the CNAME moves (the `_dstack-app-address` TXT is identical) — a single Route 53 write using the existing `CERTBOT_AWS_*` role. Runs only on a verify failure where the new box actually came up healthy (if it never came up, it never claimed and the old box still serves — no rollback needed).

So Option A is now: deploy to cold → it claims on `/health` → verify on its direct URL → on success, confirm + stop old; on failure, auto-rollback to old + stop the bad box.

Remaining recommended-but-not-blocking follow-up: deepen `/health` (codex #8 — it currently treats billing keys as informational), so the boot-time claim gate is more meaningful and the rollback fires less often. Left as a separate change since it touches app code (`health.rs`).

PR #446, opened 2026-06-04
docs(plans): private-apps-on-Lit framework + Lambda-parity gateway plan

clawdbot-glitch003:
Adds two design plans (docs only — no code changes).

`plans/private-apps-backend.md` proposes a "backend-as-Lit-Actions" framework — an in-action runtime, CLI, client SDK, and optional relay — for building private apps where Postgres row contents are encrypted at rest and decrypted only inside the TEE, with plaintext index columns for querying. `plans/chipotle-lambda-parity.md` lays out the prioritized Chipotle gateway work (per-key spend cap with auto-disable, per-key/per-IP rate + concurrency limits, origin allowlist, and a public-key type) needed to make a usage key safe to embed directly in a frontend, with blast-radius parity to a public AWS Lambda URL. Together they argue the relay becomes optional once the gateway controls land, since confidentiality is already TEE/CID-gated and only spend/availability needs bounding. Both are grounded in the existing `examples/dark-pool/` encrypted-Postgres pattern and the current `lit-api-server` guard/billing code.

### DIFF
--- a/plans/chipotle-lambda-parity.md
+++ b/plans/chipotle-lambda-parity.md
@@ -1,0 +1,212 @@
+# Chipotle: Lambda-Parity for Frontend-Callable Actions
+
+Status: design proposal / work plan
+Author: Chris (with Claude)
+Date: 2026-06-04
+Related: [private-apps-backend.md](./private-apps-backend.md) — this is the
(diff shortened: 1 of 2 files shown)


glitch003:
closing - moved plan to #477

PR #447, opened 2026-06-05
docs(billing): add auto top-up implementation plan

DashKash54:
## Summary

Design docs for the **auto top-up** feature on Stripe-paying users. No code changes; this PR establishes the architecture that subsequent implementation PRs will follow.

Two files changed:

- `plans/auto-top-up.md` — full self-contained implementation plan: architecture, Postgres schema with CHECK constraints, endpoint specs, sync trigger + sync credit flow, reconciler cron design, SCA recovery flow (on-session `stripe.confirmCardPayment`, no extra webhook), three-layer dedup defense, edge cases, explicit soft-cap trade-offs, 9-phase build plan with per-phase acceptance tests, full local development & testing setup, and a handoff checklist.
- `lit-payments/README.md` — concise auto top-up architecture section at the top: components, storage map, component-interaction diagram, left-to-right runtime flow, three-layer dedup model, and key invariants. Points to the full plan for detail.

## Architecture in one paragraph

`lit-api-server` is unchanged except for one tiny `POST /internal/invalidate_balance_cache` endpoint. The existing wallet/API-key auth (`billing_auth.rs`) is extracted into a shared crate so `lit-payments` can use it. All auto-top-up logic — dashboard endpoints, the single `customer.updated` webhook handler (which does the full trigger → sync charge → sync credit), the reconciler cron, and SCA recovery endpoints — lives in `lit-payments`. Postgres holds two new tables: `auto_topup_config` and `auto_topup_credits`.

## Left-to-right runtime flow

```
CLIENT  ─►  lit-api-server  ─►  Stripe                ─►  Stripe fires  ─►  lit-payments         ─►  Stripe                  ─►  Postgres                ─►  Stripe                  ─►  Postgres                ─►  lit-api-server
 run        deduct credits      balance_transactions     customer.updated   /stripe/webhook           paymentIntents.create       INSERT auto_topup_credits   balance_transactions         UPDATE row                 POST /internal/
 action     (existing)          write                    webhook            (verify HMAC, mutex,      (off_session, confirm)      ON CONFLICT DO NOTHING      Idempotency-Key:             SET stripe_balance_         invalidate_balance_cache
                                                                            balance fetch, list PIs,                                                          credit:{pi.id}               transaction_id              (fire-and-forget)
                                                                            cap check, failure
                                                                            derivation)
```

## Component interaction (high level)

```
USER ─► DASHBOARD ─► lit-payments        (save card, save config, SCA recovery)
                       │
                       ▼
                   Postgres auto_topup_config

CLIENT ─► lit-api-server (deduct credits — existing flow, unchanged)
                       │
                       ▼
                   Stripe balance_transactions write → customer.balance drops
                       │
                       ▼
                   Stripe fires customer.updated webhook
                       │
                       ▼
                   lit-payments POST /stripe/webhook
                       │  verify HMAC, mutex, fresh balance fetch,
                       │  list PIs (failure derivation + cap check)
                       │
                       ▼
                   Stripe paymentIntents.create (off_session, confirm) — synchronous
                       │
              ┌────────┴────────┐
              ▼ succeeded       ▼ authentication_required, declined, timeout
       INSERT auto_topup_credits   set pending state / email + banner / log
              │
              ▼
       Stripe balance_transactions (credit, Idempotency-Key: credit:{pi.id})
              │
              ▼
       UPDATE auto_topup_credits SET stripe_balance_transaction_id
              │
              ▼
       Fire-and-forget POST lit-api-server /internal/invalidate_balance_cache
              │
              ▼
       release mutex, return 200
```

## SCA recovery flow (when `authentication_required` fires)

User-driven, on-session. No extra webhook needed.

```
User                Email                  Dashboard                     lit-payments                Stripe.js                          Stripe                      Postgres
 │                    │                       │                              │                            │                                │                            │
 │ click email link   │                       │                              │                            │                                │                            │
 ├───────────────────►│                       │                              │                            │                                │                            │
 │                    │ /recover_topup?token  │                              │                            │                                │                            │
 │                    ├──────────────────────►│                              │                            │                                │                            │
 │                    │                       │ GET /billing/auto_topup_     │                            │                                │                            │
 │                    │                       │ resume?token=...             │                            │                                │                            │
 │                    │                       ├─────────────────────────────►│                            │                                │                            │
 │                    │                       │                              │ verify token, get pending_ │                                │                            │
 │                    │                       │                              │ action_pi_id ─────────────►│                                │                            │
 │                    │                       │                              │ retrieve PI ──────────────►│                                │                            │
 │                    │                       │                              │◄──── client_secret ────────│                                │                            │
 │                    │                       │◄── {client_secret} ──────────│                            │                                │                            │
 │                    │                       │ stripe.confirmCardPayment    │                            │                                │                            │
 │                    │                       │ (client_secret)              │                            │                                │                            │
 │                    │                       ├──────────────────────────────┼───────────────────────────►│                                │                            │
 │                    │                       │                              │                            │ render bank's 3DS in iframe    │                            │
 │◄── 3DS challenge in browser ───────────────┼──────────────────────────────┼────────────────────────────│                                │                            │
 │ authenticate                               │                              │                            │                                │                            │
 ├────────────────────┼───────────────────────┼──────────────────────────────┼───────────────────────────►│                                │                            │
 │                    │                       │                              │                            │ PI status = succeeded ────────►│                            │
 │                    │                       │◄── result.paymentIntent ─────┼────────────────────────────│                                │                            │
 │                    │                       │   .status='succeeded'        │                            │                                │                            │
 │                    │                       │ POST /billing/auto_topup_    │                            │                                │                            │
 │                    │                       │ resume/complete              │                            │                                │                            │
 │                    │                       ├─────────────────────────────►│                            │                                │                            │
 │                    │                       │                              │ verify PI via GET PI ─────►│                                │                            │
 │                    │                       │                              │ INSERT auto_topup_credits ─┼────────────────────────────────┼───────────────────────────►│
 │                    │                       │                              │ balance_transactions       │                                │                            │
 │                    │                       │                              │ (Idempotency-Key:          │                                │                            │
 │                    │                       │                              │  credit:{pi.id}) ─────────►│                                │                            │
 │                    │                       │                              │ UPDATE row ────────────────┼────────────────────────────────┼───────────────────────────►│
 │                    │                       │                              │ clear pending + token ─────┼────────────────────────────────┼───────────────────────────►│
 │                    │                       │                              │ invalidate lit-api-server  │                                │                            │
 │                    │                       │                              │ cache (fire-and-forget)    │                                │                            │
 │                    │                       │◄── 200 OK ───────────────────│                            │                                │                            │
 │                    │                       │ show "topped up!"            │                            │                                │                            │
 │◄───────────────────┴───────────────────────│                              │                            │                                │                            │
```

## Reconciler cron (every 15 minutes)

Recovers from HTTP timeouts and partial writes. Load-bearing in the sync-credit design.

```
lit-payments cron tick (every RECONCILER_INTERVAL_SECS, default 900)
       │
       ▼
Postgres SELECT customer_id FROM auto_topup_config WHERE enabled
       │
       ▼ for each customer:
Stripe GET /v1/payment_intents?customer=...&created[gte]=now-7d
       filter to metadata.source=auto_topup AND status=succeeded
       │
       ▼ for each succeeded PI:
Postgres SELECT FROM auto_topup_credits WHERE payment_intent_id = pi.id
       │
       ├─ row + balance_transaction_id populated → skip (already done)
       ├─ row but balance_transaction_id NULL → retry credit (idempotent), UPDATE row
       └─ no row → run full credit (INSERT row, balance_transactions, UPDATE row), alert Sally
```

## Key design decisions (validated independently)

- **Single webhook subscription** (`customer.updated`) — empirically verified 1:1 firing per `balance_transactions` write, no coalescing.
- **Sync charge + sync credit** in the same handler — no `payment_intent.succeeded` webhook needed for the happy path.
- **Reconciler cron (15-min default)** is load-bearing — recovers HTTP timeouts and partial writes.
- **SCA support via on-session recovery flow** — tokenized email link + dashboard `stripe.confirmCardPayment`; supports EU/SCA cards without subscribing to `payment_intent.succeeded`.
- **Soft cap** — rare list-endpoint staleness can overshoot by ~1 top-up; UI says "approximately $X/month."
- **Bias toward more top-ups, never fewer** — extra charge over-credits the user's account (recoverable); missed top-up fails a Lit Action mid-flight (user-visible).

Independently validated by Codex consult across four review passes plus an empirical Stripe test-mode webhook firing experiment.

## Test plan

- [X] Product review of soft-cap copy ("approximately $X/month")
- [ ] Confirm minimum top-up amount, failure threshold, and bucket window with product
- [X] Confirm 15-min reconciler interval is acceptable for support / UX
- [X] Confirm SCA recovery email template wording and recovery-page UX

### DIFF
--- a/plans/auto-top-up.md
+++ b/plans/auto-top-up.md
@@ -2,69 +2,74 @@
 
 **Status:** Locked. Ready to implement.
 **Scope:** Stripe-paying users only. LITKEY / crypto users are out of scope.
-**Repository:** `chipotle` (this repo).
+**Architecture:** Webhook-driven, lit-payments-owned. lit-api-server is unchanged except for one tiny cache-invalidation endpoint.
 
 ---
 
 ## 1. Goal
 
-After every Lit Action deduction (and every management deduction), if the user's available Stripe credit falls below a threshold they configured, automatically charge their saved card off-session for a configured top-up amount, subject to a configured monthly cap. UI is modeled on Claude Console and OpenAI Platform's auto-recharge modals.
+After any Lit Action or management deduction reduces a user's Stripe customer balance below a configured threshold, automatically charge their saved card off-session for a configured top-up amount, subject to a configured monthly cap. UI matches Claude Console and OpenAI Platform's auto-recharge modals.
 
-**Auto top-up is opt-in.** Users configure it via the dashboard after saving a card.
+Auto top-up is opt-in. Users configure it via the dashboard after saving a card.
 
 ---
 
 ## 2. Product framing
 
-### What the user sees
-- A new "Auto top-up" section in the dashboard's billing UI.
-- A modal with: enable toggle, threshold input, top-up amount input, monthly cap input, saved-card picker.
-- Status banner: shows current rule in plain English when enabled, or a "set up a card" CTA when disabled.
-- Failure banner: shown when auto-top-up is auto-disabled after 3 consecutive failures, or when the last off-session attempt requires SCA re-authentication.
+### Soft cap, explicit trade-off
 
-### Soft cap (explicit trade-off)
-The monthly cap is **best-effort**, not a contract. UI copy must say "approximately $X/month" or "up to ~$X/month with rare overage." Overshoot is bounded by one top-up amount at 5-minute idempotency-key bucket boundaries.
+Monthly cap is **best-effort**, not a hard contract. UI copy must say "approximately $X/month." The only way the cap can be exceeded is a rare race against Stripe's PaymentIntent list endpoint, which isn't strongly read-after-write consistent. Probability <1% per opportunity; overshoot bounded by one top-up amount when it fires.
+
+### Bias toward more top-ups, never fewer
+
+A missed top-up means Lit Action execution can fail mid-flight when balance hits zero. That's user-visible and disastrous. An extra top-up means the user has unexpected credits in their account — recoverable via refund through Sally's admin portal. The design intentionally favors over-charging over under-charging in any ambiguous case.
 
 ### What we are not building
+
 - Hard monthly cap with reservation ledger.
 - Currency support beyond USD.
 - Refund-aware cap accounting (refunds do NOT restore monthly cap capacity).
 - Automated daily/rolling restrictions for public-tier API keys (future feature).
-- Reconciler cron — manual recovery via Sally's admin portal for the rare "webhook lost for 3 days" case.
+- Reconciler cron — Stripe's 3-day webhook retry + manual recovery via admin portal cover the rare delivery-failure case.
 
 ---
 
 ## 3. Architecture overview
 
+### Trigger source: Stripe `customer.updated` webhook
+
+When `lit-api-server` deducts credits via `POST /v1/customers/{id}/balance_transactions`, Stripe fires a `customer.updated` event containing the new balance and the previous balance. `lit-payments` listens to this event, filters for balance changes, and decides whether to top up.
+
+**Empirically validated** in Stripe test mode (June 2026): 21 balance_transactions → 21 `customer.updated` events, 1:1, no coalescing under either slow (0.5s gap) or rapid parallel (10 in 2.8s) load. Each event contains the new `balance` and `previous_attributes.balance` for delta detection.
+
 ### Components
 
 | Component | Path / Location | Role |
 |---|---|---|
-| **Dashboard** | `lit-static/dapps/dashboard/` | Vanilla HTML/JS. Adds config modal, save-card flow, status banners. |
-| **lit-api-server** | `lit-api-server/`, runs in TEE | Existing API server. Adds: 4 new dashboard-facing endpoints behind existing `billing_auth.rs`, 1 fire-and-forget trigger call after every deduction, 1 internal cache-invalidation endpoint. |
-| **lit-payments** | `lit-payments/`, on Railway | Existing service. Adds: 3 internal-only endpoints, 1 Stripe webhook receiver, 2 Postgres tables. |
-| **Postgres** | inside `lit-payments` DB | Existing DB. Adds 2 new tables. |
-| **Stripe** | external | Customer, PaymentMethod, PaymentIntent, balance transactions, webhooks. |
+| **Dashboard** | `lit-static/dapps/dashboard/` | Vanilla HTML/JS. Adds config modal, save-card flow, status banners. Talks directly to `lit-payments`. |
+| **lit-api-server** | `lit-api-server/`, runs in TEE | Existing API server. **No trigger code, no auto-top-up logic.** Only adds: 1 internal `invalidate_balance_cache` endpoint. The existing wallet/API-key auth module (`billing_auth.rs`) is extracted to a shared crate so `lit-payments` can use it. |
+| **lit-payments** | `lit-payments/`, on Railway | Existing service. Adds: 3 dashboard-facing endpoints (now with wallet/API-key auth), 2 Stripe webhook handlers, 2 new Postgres tables. Becomes the home of all auto-top-up logic. |
+| **Postgres** | inside `lit-payments` DB | Existing DB. Adds 2 new tables: `auto_topup_config`, `auto_topup_credits`. |
+| **Stripe** | external | Customer, PaymentMethod, PaymentIntents, balance transactions, webhooks (`customer.updated`, `payment_intent.succeeded`, `payment_intent.payment_failed`). |
+| **Shared auth crate** | new, e.g. `lit-billing-auth/` | Extracted from `lit-api-server::core::v1::guards::billing_auth`. Accepts wallet signature (EIP-712) and API key. Used by both lit-api-server (existing endpoints) and lit-payments (new dashboard endpoints). |
 
-### Auth model
-- Dashboard sends the same headers it already sends today: either an API key OR a wallet signature (EIP-712 ChainSecured).
-- `lit-api-server`'s existing `billing_auth.rs` guard (`src/core/v1/guards/billing_auth.rs`) verifies both and derives the Stripe `customer_id`.
-- `lit-payments` is internal-only. Its endpoints accept only:
-  - `X-Internal-Secret` from `lit-api-server` (high-entropy shared secret, TLS-only, constant-time compare, never logged).
-  - `Stripe-Signature` HMAC on the webhook endpoint.
+### Why the TEE no longer holds auto-top-up logic
+
+Architectural principle: keep the TEE (lit-api-server) narrowly focused on Lit Action execution and key-usage operations. Anything that can live outside the TEE should. Auto-top-up logic doesn't require a TEE, so it lives in `lit-payments`. The empirical test confirms Stripe's `customer.updated` webhook reliably delivers per-balance-change events, so we don't need an internal trigger from lit-api-server.
 
 ### Storage map
 
-| Data | Where it lives | Why |
+| Data | Location | Reason |
 |---|---|---|
-| Config (`enabled`, `threshold_cents`, `topup_amount_cents`, `monthly_cap_cents`, `payment_method_id`) | Postgres `auto_topup_config` | Fast reads (~1ms), atomic writes, schema evolution, queryable for support |
-| Consent record (`consent_version`, `consent_signed_at`) | Postgres `auto_topup_config` (same row) | Off-session merchant-initiated charges require recorded user consent |
-| Card data | Stripe (PaymentMethod attached to Customer) | Never touches our servers; PCI scope stays with Stripe |
-| Wallet ↔ Stripe Customer mapping | Stripe customer metadata (`metadata.wallet_address`) | Existing pattern; unchanged |
-| Charge history (PaymentIntents) | Stripe (filtered by `metadata.source=auto_topup`) | Source of truth for what was charged |
-| Monthly spend total | Computed by listing Stripe PIs | No counter to race on |
-| Failure state | Derived by listing last N PIs and counting failures | No counter to race on |
-| Credit dedup (1 row per credited PI) | Postgres `auto_topup_credits` | Permanent dedup beyond Stripe's 24h idempotency cache |
+| Config (5 fields: `enabled`, `threshold_cents`, `topup_amount_cents`, `monthly_cap_cents`, `payment_method_id`) | Postgres `auto_topup_config` | Fast reads, atomic writes, schema evolution, admin queryability |
+| Consent record (`consent_version`, `consent_signed_at`) | Postgres `auto_topup_config` (same row) | Off-session merchant-initiated charges require recorded user consent (PSD2 / SCA + Stripe policy) |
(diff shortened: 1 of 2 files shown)


plans/auto-top-up.md
```diff
@@ -0,0 +1,882 @@
...
+## 2. Product framing
+
+### What the user sees
+- A new "Auto top-up" section in the dashboard's billing UI.
+- A modal with: enable toggle, threshold input, top-up amount input, monthly cap input, saved-card picker.
+- Status banner: shows current rule in plain English when enabled, or a "set up a card" CTA when disabled.
+- Failure banner: shown when auto-top-up is auto-disabled after 3 consecutive failures, or when the last off-session attempt requires SCA re-authentication.
+
+### Soft cap (explicit trade-off)
+The monthly cap is **best-effort**, not a contract. UI copy must say "approximately $X/month" or "up to ~$X/month with rare overage." Overshoot is bounded by one top-up amount at 5-minute idempotency-key bucket boundaries.
```
glitch003: why is this the case?  2 things: 
1. we should be able to guarantee we don't go over the cap.  like the design should have no race conditions or ways to go over the cap.  you have a per-customer mutex - within that mutex it should be possible to just check if we're at the cap and not charge them if we are.  like can you explain how we could go past the cap?
2. why is there a 5 min idempotency window?  does that not mean that if the user needs to top up twice in 5 mins, they can't?

@DashKash54 you've seen the screenshots of anthropic and openai's UI here - they do not say "approximately" or "with rare overage" so why would we have to say that?  is there something unique about our architecture that makes it impossible to make this guarantee like anthropic and openai do?  i don't think there are.

plans/auto-top-up.md
```diff
@@ -0,0 +1,882 @@
...
+
+---
+
+## 5. Endpoints
+
+### Dashboard-facing (on `lit-api-server`, behind existing `billing_auth.rs`)
+
+| Method + Path | Purpose | Behavior |
+|---|---|---|
+| `POST /billing/setup_intent` | Save card | Calls `lit_billing_core::customer::find_or_create_by_wallet`. Refuses if user has no Stripe customer AND has never made a manual top-up (we don't auto-create customers here; user must do their first manual top-up first to bootstrap the Stripe customer). Then creates Stripe SetupIntent via `lit_billing_core::StripeClient::post_with_idempotency("setup_intents", [usage=off_session, customer=cus_xxx])`. Returns `client_secret` to dashboard. |
```
glitch003: does this really need to go through the TEE?  why wouldn't you just hit lit-payments directly?

plans/auto-top-up.md
```diff
@@ -0,0 +1,882 @@
...
+| `GET /billing/auto_topup_config` | Read config | Forwards to `lit-payments`'s `GET /internal/auto_topup_config?customer_id=...` with `X-Internal-Secret`. Returns config JSON to dashboard. |
```
glitch003: same question as above.  this endpoint literally just forwards to lit-payments.  why not just hit lit-payments directly?

plans/auto-top-up.md
```diff
@@ -0,0 +1,882 @@
...
+| `PUT /billing/auto_topup_config` | Save config | Body: `{enabled, threshold_cents, topup_amount_cents, monthly_cap_cents, payment_method_id, consent_version}`. Server-side validation: verify `payment_method_id` belongs to this customer (`GET /v1/customers/{cus_xxx}/payment_methods` and check membership), enforce `cap >= topup_amount`, etc. Forwards to `lit-payments`'s `PUT /internal/auto_topup_config`. |
```
glitch003: again, makes no sense in my mind to send this through the TEE as well

plans/auto-top-up.md
```diff
@@ -0,0 +1,882 @@
...
+| `POST /internal/invalidate_balance_cache` | Drop cached balance | Called by `lit-payments` after every successful auto-credit. Body: `{customer_id}`. Calls `state.balance_cache.invalidate(&customer_id)` (same primitive as existing `confirm_payment_intent` flow at `lit-api-server/src/stripe.rs:612`). Auth: `X-Internal-Secret`. |
```
glitch003: IMO this is the only actual endpoint we need to add to lit-api-server

plans/auto-top-up.md
```diff
@@ -0,0 +1,882 @@
...
+      amount: topup_amount_cents
+      currency: usd
+      off_session: true
+      confirm: true
+      metadata[source]: auto_topup
+      metadata[wallet_address]: 0x...
+      Idempotency-Key: {key}
+    ```
+    On `authentication_required` error: set a flag on the config row (e.g. via `disabled_reason='requires_action'`) so the dashboard shows "action required" banner. User must re-authenticate on-session next visit.
+    On `card_declined` / `expired_card` / etc.: webhook will fire `payment_intent.payment_failed` and trigger the counter-derivation in step 6 next time. No write here.
```
glitch003: we should tell the user in an email if this happens, not just "no write here"

plans/auto-top-up.md
```diff
@@ -0,0 +1,882 @@
...
+      payment_method: pm_xxx (from config)
+      amount: topup_amount_cents
+      currency: usd
+      off_session: true
+      confirm: true
+      metadata[source]: auto_topup
+      metadata[wallet_address]: 0x...
+      Idempotency-Key: {key}
+    ```
+    On `authentication_required` error: set a flag on the config row (e.g. via `disabled_reason='requires_action'`) so the dashboard shows "action required" banner. User must re-authenticate on-session next visit.
```
glitch003: is the dashboard's ability to resolve this part of this plan?  i don't see it.  we should probably also email the user.

glitch003 (changes requested):
thinking about this more...  i don't think we need any changes to lit-api-server except for the invalidate cache endpoint.  

imo the cleanest model is to not involve lit-api-server at all, and just use lit-payments directly.  we can use stripe webhooks that hit lit-payments when the user's credit balance changes.  i don't see any issues with that model but maybe there's something i'm missing so please let me know if you see any issues.
so the sequence would be

1. user sets up auto top up with endpoint on lit-payments.  they set the card, the config, etc.
2. user is using chipotle.  stripe is sending webhooks to lit-payments every time their balance changes.
3. the webhook fires.  look at the balance in the webhook data.  if their balance is above the auto top up config, just exit. if it's below the auto top up config, then grab a mutex on the customer.  it may be time for an auto top up, so let's do a further investigation to see if we should top them up, locked with the mutex so that this process can only run once at a time per customer.
4.  check their current balance on stripe (don't blindly trust the webhook data, it might be old data).  check if this balance is below the auto top up config.  if it is, charge and credit them, and release the mutex.

DashKash54:
> imo the cleanest model is to not involve lit-api-server at all, and just use lit-payments directly.

The whole reason we involve the lit-api-server in this instead of directly hooking up the lit-payments is because lit-payments doesn't have any user-level auth. The three dashboard-facing endpoints (save card, save config, read config) are per-user actions; without auth anyone could read another user’s auto-top-up config or worse or write to it (attack — disable someone’s auto-top-up, change their config).

Where auth lives today: only on lit-api-server (billing_auth.rs, supporting both wallet signature and API key). Therefore, putting dashboard endpoints on lit-payments forces one of:
1. Duplicate billing_auth.rs into lit-payments → two sources of truth, drift risk, double maintenance.
2. Extract billing_auth.rs into a shared crate → real refactor, broader review surface, more PR scope.
3. Build new auth on lit-payments → worst option.

Recommended approach: Keeping dashboard endpoints on lit-api-server (current design): zero auth changes. lit-api-server’s existing guard handles everything. lit-payments only ever sees service-authed requests with the customer_id already derived. That’s why we route the dashboard through lit-api-server and keep lit-payments internal.

DashKash54:
> we can use stripe webhooks that hit lit-payments when the user's credit balance changes.

The Stripe events catalog at https://docs.stripe.com/api/events/types is the exhaustive list of webhook events Stripe ships. The event the proposal needs — a per-balance-change webhook for customer.balance — is not on it. The events that ARE on it (customer_cash_balance_transaction.created, cash_balance.funds_available) cover a different Stripe primitive (customer.cash_balance, the bank-transfer-funded balance) that Lit doesn’t use. The Customer Balance Transactions API reference at https://docs.stripe.com/api/customer_balance_transactions also has no webhook section, unlike every other major Stripe API. So the “Stripe sends webhooks every time balance changes” assumption doesn't seem to be true for our primitive — the only catch-all available is customer.updated, which fires on any customer field change and gives you no delta. To make the proposal work we need to migrate off customer.balance to a different primitive (which is much more unnecessary work for the auto top-up feature).

plans/auto-top-up.md
DashKash54: Because lit-payments doesn't have its own user-level auth so either we can duplicate from lit-api-server: wallet/EIP-712 verification, API key validation, customer lookup. Real work, real maintenance burden, second source of truth for auth. Or build new auth on lit-payments.

All we're doing is saving setup info and reading so routing dashboard endpoints through lit-api-server adds essentially very little load to lit-api-server, and zero load on the Lit Action hot path

plans/auto-top-up.md
DashKash54: Same reason as above

plans/auto-top-up.md
DashKash54: Same reason as above

plans/auto-top-up.md
DashKash54: Yeah we def need this invalidation

plans/auto-top-up.md
DashKash54: Yeah the goal is to send an email + dashboard banner as mentioned in Phase 7: "Your auto top-up of $X failed: {reason}. Please update your card."

plans/auto-top-up.md
DashKash54: Good catch added more details in the plan to handle this

plans/auto-top-up.md
DashKash54: 1. We can’t hard-guarantee because we’re using Stripe’s list endpoint to compute the current month’s auto-top-up spend, and list endpoints only give eventual consistency (unlike direct GET-by-id, which is read-after-write consistent). So if a user is operating close to the cap and trigger A gets the mutex, charges via Stripe, and releases — when trigger B then acquires the mutex and calls Stripe’s list endpoint, the response can still be stale and miss A’s just-created PaymentIntent. B undercounts the month’s spend, passes the cap check, and creates another charge. Low probability but real. The only way to hard-guarantee the cap is to maintain a local ledger (Postgres row per charge under a row lock or advisory lock) so the cap sum doesn’t depend on Stripe’s list freshness — which adds real complexity I deliberately deferred for the soft-cap product trade-off since it's just an additional top-up which firstly is less likely to happen and can be refunded manually if required

DashKash54: 2. The prob of the Stripe list being stable is low though, so we can just skip it and and accept the occasional double-charge. This ensure that a requested top-up always happens (even within the 5 min window) with the occasional second top-up. As to hard-guarantee we need a local ledger

DashKash54:
Closing since the architecture has changed quite a bit in PR 448

PR #448, opened 2026-06-08
feat(auto-topup): Stripe auto top-up for opted-in users

DashKash54:
# Auto top-up — Stripe-paying users

Adds opt-in **automatic credit top-ups** to Lit Protocol. When a user's Stripe customer balance drops below a configured threshold, we charge their saved card off-session and credit their Lit-action balance — without the user re-authenticating, except for European SCA-required cards (where we send a one-time recovery email).

This is one PR (~13k lines, one feature). There are no dependent PRs — the architectural follow-up (#466) was merged into this branch and is now part of this same diff.

---

## What the system does

1. User saves a card on the dashboard (Stripe Elements, 3DS at save time for SCA cards — creates MIT prior-auth).
2. User sets thresholds: "drops below $X, restore to $Y, max $Z/month."
3. Every Lit Action / management deduction calls Stripe `balance_transactions`, which fires a `customer.updated` webhook to lit-payments.
4. The webhook handler runs the full charge + credit pipeline **synchronously**: verifies signature, mutex, fresh balance fetch, PI list, failure derivation, cap check, off-session `paymentIntents.create`, `balance_transactions` write, cache invalidation. Returns 200 only after the credit lands.
5. If Stripe returns `authentication_required` (SCA), we stage a recovery token, email the user with a 24h one-time link, and pause auto top-up for that customer. The recovery page calls `stripe.handleNextAction()` to trigger the bank's 3DS challenge on the existing PI. On success, the same sync-credit path applies.
6. A reconciler tokio task sweeps every 15 minutes: any partial credit row in the DB (`stripe_balance_transaction_id IS NULL`) is repaired by replaying the balance_transactions write with the same idempotency key. Catches HTTP timeouts and mid-flow crashes.

---

## Architecture

```
                                       ┌──────────────────┐
                                       │   Stripe (test/  │
                                       │   prod modes)    │
                                       │   - Customers    │
                                       │   - PaymentMethod│
                                       │   - PaymentIntent│
                                       │   - BalanceTxn   │
                                       └────────▲─────────┘
                                                │
                                       customer.updated
                                                │
┌─────────────────────────┐          ┌─────────▼────────────┐
│  Dashboard              │          │  lit-payments         │
│  (lit-static/dapps/     │  PUT     │  (Railway / Fly)      │
│   dashboard/)           │  GET     │                       │
│  - Auto recharge modal  │◄────────►│  POST /stripe/webhook │
│  - SCA recovery page    │          │  GET/PUT /billing/*   │
│  - Status banner        │          │  Reconciler (tokio)   │
└────────────┬────────────┘          │  Resend (email)       │
             │                       │  Postgres             │
             │ wallet-sig auth       │   - auto_topup_config │
             │ (X-Wallet-Auth)       │   - auto_topup_credits│
             │                       └─────────┬─────────────┘
             ▼                                 │
     ┌────────────────────┐                    │ /internal/
     │ lit-api-server     │ ◄──────────────────┘ invalidate_balance_cache
     │ (TEE)              │   (cache invalidation only — no auth hops)
     │  - Lit Actions     │
     │  - Customer mgmt   │
     │  - Balance cache   │
     └────────────────────┘
```

### Where each piece of data lives

| Data | Where | Why |
|------|-------|-----|
| Auto-topup rule (enabled, threshold, top-up, cap, payment_method_id) | Postgres `auto_topup_config` | Fast reads, atomic writes, schema evolution |
| Consent (version + signed_at) | Same row | Off-session merchant-initiated charging requires recorded consent |
| Wallet ↔ Stripe customer mapping | Stripe customer metadata (`metadata.wallet_address`) | Existing pattern, unchanged |
| Card data | Stripe (PaymentMethod, attached to Customer) | PCI stays with Stripe — we never see PANs |
| Charge history | Stripe PaymentIntents (`metadata.source=auto_topup`) | Source of truth |
| Monthly spend total | Computed on demand by listing PIs | No counter to race on |
| Failure state | Derived by listing recent PIs and counting consecutive failures | No counter to race on |
| Credit dedup ledger (1 row per credited PI) | Postgres `auto_topup_credits` | Reconciliation primitive — finds partial / orphan credits |
| SCA recovery token | Postgres `auto_topup_config` (same row) | 24h one-time bearer; tied to the pending PI |

### Concurrency safety (three layers)

| Layer | Prevents | How | Scope |
|-------|----------|-----|-------|
| In-process per-customer mutex (moka cache, 5-min TTL) | Parallel `customer.updated` events for the same customer doing duplicate evaluations | `Cache<String, Arc<tokio::sync::Mutex<()>>>` | Per-process |
| Postgres `UNIQUE(payment_intent_id)` on `auto_topup_credits` | Double-credit (sync handler + reconciler both trying same PI) | `INSERT ... ON CONFLICT DO NOTHING` | Permanent |
| Stripe Idempotency-Key on `balance_transactions` write | Double-credit from retry under transient Stripe/network failure | `credit:{pi.id}` | Stripe-side, 24h |

Layers 2 and 3 are the correctness primitives. The mutex is an optimization.

---

## Components

| Component | Path | Owns |
|-----------|------|------|
| **lit-payments** | `lit-payments/` | All auto-top-up logic: dashboard endpoints, webhook handler, sync charge + sync credit, reconciler, SCA resume |
| **lit-billing-core** | `lit-billing-core/` | Shared infra: Stripe HTTP client, customer/balance helpers, EIP-712 verifier, on-chain api-key resolver, `billing_auth` module (BillingAuth guard + AuthResolver trait) |
| **lit-api-server** | `lit-api-server/` | Unchanged Lit Actions path. Adds one endpoint: `POST /internal/invalidate_balance_cache` for the post-credit cache flush |
| **Dashboard** | `lit-static/dapps/dashboard/` | Auto-recharge modal, save-card sub-flow, SCA recovery page, status banners |
| **Postgres** | `lit-payments/migrations/` | Two new tables (`auto_topup_config`, `auto_topup_credits`) with CHECK constraints + partial indexes |

### Shared crate extraction (was the glitch architectural follow-up, now merged into this PR)

Originally Phase 2 created a separate `lit-billing-auth` crate that wrapped the auth surface, and `lit-payments` HTTP-hopped to `lit-api-server` for both EIP-712 wallet-signature verification (`/internal/verify_wallet_auth`) and API-key → wallet resolution (`/internal/resolve_api_key`). PR reviewer (glitch) correctly pointed out that:

1. EIP-712 signature recovery is a pure function — no TEE-bound state, no reason for an HTTP hop
2. The on-chain api-key resolver just reads a public contract mapping — `lit-payments` can read it directly with its own RPC URL
3. With only two consumers, a separate crate added overhead without real isolation

**Resolution** (commits `e19e8006..e5e93671`):
- EIP-712 verifier moved into `lit-billing-core::eip712`
- On-chain API-key resolver moved into `lit-billing-core::on_chain` (configurable via env: RPC URL + contract address + chain ID)
- `lit-billing-auth` crate folded into `lit-billing-core::billing_auth` module; crate deleted
- `lit-payments` swapped its `HttpAuthResolver` for `LocalAuthResolver` that calls the moved functions in-process
- `/internal/verify_wallet_auth` and `/internal/resolve_api_key` endpoints removed from `lit-api-server` (`/internal/invalidate_balance_cache` kept — that one IS a genuine cross-service call)

**Net result**: one less HTTP hop per authenticated request, one less shared-secret round-trip surface for auth, one less crate in the workspace. The `X-Internal-Secret` header is still used for the legit cache-invalidation call.

---

### Dashboard-facing (`lit-payments`, behind `BillingAuth` guard — wallet-sig or API-key)

| Method | Path | Purpose |
|--------|------|---------|
| `POST` | `/billing/setup_intent` | Returns SetupIntent client_secret + publishable key for save-card flow |
| `GET` | `/billing/auto_topup_config` | Read config + brand/last4 of saved card (Stripe-enriched) |
| `PUT` | `/billing/auto_topup_config` | Save config — validates pm ownership, $5 floor, $200 ceiling, cap (optional) ≥ topup |
| `GET` | `/billing/payment_method?pm_id=...` | Just-saved-card brand/last4 lookup, ownership-checked |
| `GET` | `/billing/auto_topup_resume?token=...` | SCA recovery page entry; returns PI client_secret (unauthenticated by guard — token IS the credential) |
| `POST` | `/billing/auto_topup_resume/complete` | After 3DS succeeds in browser, applies the sync credit |

### Webhook (`lit-payments`, HMAC-signed by Stripe)

| Method | Path | Purpose |
|--------|------|---------|
| `POST` | `/stripe/webhook` | Single event type: `customer.updated`. Filtered by `previous_attributes.balance`. The trigger for the entire auto-topup pipeline. |

### Internal (`lit-api-server`, behind `X-Internal-Secret` constant-time compare)

| Method | Path | Purpose |
|--------|------|---------|
| `POST` | `/internal/invalidate_balance_cache` | Called by lit-payments after sync credit lands. Body: `{customer_id}` |

---

### Automated (in this PR)

| Crate | Test count | Coverage |
|-------|-----------|----------|
| `lit-payments` (lib) | 95 | Webhook signature verification, mutex serialization, sync-credit happy paths, decline + auto-disable, SCA staging, reconciler partial repair, DB upsert idempotency, config validation, internal-secret constant-time compare |
| `lit-billing-core` (lib) | 61 | EIP-712 verifier (8 primary types), Stripe HTTP error parsing, on-chain resolver selector |
| `lit-api-server` (lib) | 198 | All pre-existing tests still pass after the auth-extraction refactor |

Integration tests against real Stripe test mode are env-guarded (`STRIPE_SECRET_KEY` + `DATABASE_URL`). They silent-skip when the env is missing, so CI without credentials still goes green.

### Manual end-to-end (real Stripe test mode + real Resend + real Postgres)

Done in browser against your own Stripe test account (`acct_1GPhYrGhjEGDNSRy`) using `stripe listen` to forward webhooks to local lit-payments, the dashboard served from `lit-static/dapps/dashboard/`, and an env-gated `LIT_DEV_WALLET_BYPASS` to simulate a logged-in wallet without spinning up the TEE-only `lit-api-server`.

**Test 1 — happy path** ✅
1. Save card (`4242 4242 4242 4242`) via the auto-recharge modal — Stripe Elements collects card details inside its iframe; backend never sees card data
2. Enable auto-topup (threshold $5, restore $15, cap $100), Save
3. Drop the Stripe customer's credit balance via the Stripe Dashboard "Credit balance adjustment" widget → fires `customer.updated`
4. Within ~5 seconds: PI of $15 created and succeeded at Stripe, `auto_topup_credits` row written with `stripe_balance_transaction_id` populated, customer balance updated from `−$1` to `−$16`
5. Quick-exit verified separately: adjusting balance _above_ threshold returns 200 immediately, no PI created, no new DB row

**Test 2 — SCA / 3DS card** ✅
1. Save card (`4000 0027 6000 3184` — requires 3DS on every charge)
2. Enable auto-topup, drop customer balance below threshold
3. Webhook fires → `paymentIntents.create` off-session returns `authentication_required` (Stripe can't do 3DS off-session, so the PI moves directly to `requires_payment_method` with the error code)
4. Handler stages `pending_action_pi_id` + generates 24h recovery token + dispatches Resend email
5. Email arrives with **amount, "Visa •••• 4242", "Lit Protocol" merchant, styled button** — clicking it loads `/recover_topup.html?token=…`
6. Recovery page calls `GET /billing/auto_topup_resume?token=…` → backend returns `client_secret` + saved `pm_id` (the pm_id is detached from the PI at Stripe after the off-session failure, so frontend has to re-attach explicitly)
7. User clicks "Confirm" → `stripe.confirmCardPayment(client_secret, { payment_method: pm_id })` re-confirms the PI on-session → bank's 3DS challenge popup appears
8. User completes 3DS → PI moves through `requires_action` → `succeeded` → page shows "Topped up! Your account credit is updated"
9. Backend `/complete` endpoint applies the credit through the same sync path as Test 1

**Test 3 — auto-disable after 3 consecutive failures** ✅
1. Multiple back-to-back `customer.updated` events on a customer whose card can't off-session-confirm (the SCA test card during Test 2's exploration period)
2. Handler's failure derivation walked Stripe's PI list, counted 3+ consecutive failures, called `db::disable_after_failures`
3. Dashboard banner flipped to red **"Auto recharge paused — Three consecutive charges failed. Update your card and re-enable."** with Manage button

**Edge cases tested in flight, all fixed in-PR:**
- Stale `pending_action_pi_id` whose PI moved out of `requires_action` at Stripe — handler now fetches PI status and clears pending if it's no longer `requires_action`
- Old broken JS that mounted a Payment Element on the recovery page — replaced with `confirmCardPayment(secret, { payment_method })`
- Token consumed but pending state lingered — escape hatch detects and unwinds
- 24h-expired recovery token — same escape hatch (time-based + Stripe-status-based)
- `lit-payments` env vars after the post-glitch refactor merge (`LIT_ACCOUNTS_RPC_URL` / `CHAIN_ID` / `CONTRACT_ADDRESS`) — placeholders documented for dev; required in prod

---

## Codex review history

This PR went through 5 rounds of independent codex CLI review:

1. Initial per-phase review on the original 7-phase combined branch — 5 findings, P1s folded into the relevant phase commits
2. Per-phase re-review after the first P1 fix round — surfaced 11 P1 items, addressed in `b9e164a7`
3. Glitch architectural review — addressed in `e19e8006..e5e93671` (the now-merged follow-up)
4. Final per-phase review on the post-fix tip — 0 P0, 3 unique P1s (extreme edge cases, deferred per scope decision)
5. Focused review on the architectural refactor — 1 P1 caught: wrong ABI selector (`bytes32` vs `uint256`) on the on-chain resolver, fixed in `ee192965`

Per-phase codex comments are posted on the PR; consolidated summary at https://github.com/LIT-Protocol/chipotle/pull/448#issuecomment-4657364314.

---

## What's NOT in this PR (intentional scope decisions)

- **Decline / auto-disable email templates**: only the SCA recovery email is wired. Decline notifications and auto-disable-after-3-failures emails are deferred — failure state is fully derived from the PI history and the dashboard shows the right banner, but the Resend dispatch for those two paths is a follow-up.
- **Metrics instrumentation**: trigger count, charge success/failure rate, reconciler activity — TODO for ops hardening.
- **Refund-aware monthly cap**: refunds don't restore cap budget. Documented trade-off.
- **Currency other than USD**: out of scope.
- **Public-tier API-key rolling restrictions**: out of scope.

---

🤖 Built with [Claude Code](https://claude.com/claude-code)


DashKash54:
## Codex review — Phases 1–7

Independent review by GPT-5 codex. Asked for severity-ranked findings with file:line citations against the locked plan + per-phase context.

### Findings

- **[P1] SCA recovery loses the PaymentIntent id** — `lit-billing-core/src/http.rs:21,42` collapses Stripe error responses to a formatted string, discarding `error.payment_intent.id`. The webhook handler (`lit-payments/src/auto_topup/webhook/handler.rs:273`) then regexes that string with `extract_pi_id` and, on a miss, stores the entire error message in `pending_action_pi_id` (`handler.rs:482`). Real `authentication_required` recovery would call Stripe with a non-PI id and 5xx forever.

- **[P1] Pending SCA state does not pause auto top-up** — `set_pending_action` (`lit-payments/src/auto_topup/db.rs:182,197`) writes `disabled_reason = 'requires_action'` but leaves `enabled = true`. `process_event` at `handler.rs:161` only short-circuits on `!enabled`, so the next balance-changing webhook can spawn a second off-session PI, overwrite `pending_action_pi_id`, and burn the recovery token while the first SCA action is still pending.

- **[P2] Recovery token burned before Stripe retrieval succeeds** — `consume_recovery_token` (`lit-payments/src/auto_topup/db.rs:341`) atomically clears the token, *then* `get_auto_topup_resume` (`lit-payments/src/billing/sca_resume.rs:74,94`) calls Stripe. Any transient Stripe failure surfaces as 503 and the user's second click yields 404 with no recovery path.

- **[P2] Payment-method ownership check silently ignores pagination** — `verify_payment_method_owned` (`lit-payments/src/billing/auto_topup_config.rs:110,122`) fetches one page of 100 cards. A legitimately-owned `pm_xxx` beyond page 1 is rejected. Comment in code says "paginated"; code does not.

- **[P3] `_authping` placeholder route still mounted publicly** — `lit-payments/src/main.rs:65,90`. Marked throwaway in Phase 2 commit message but never removed. Either delete or hide behind `#[cfg(test)]`.

### Defended choices (codex acknowledged correct)

- Wallet-sig only with 501 for API-key callers in `setup_intent.rs:56` + `auto_topup_config.rs:62` matches locked phase scope.
- Raw `Data` body + HMAC in `handler.rs:58` is correct; `Json<>` extractor would break signature verification.
- Fresh balance fetch *after* mutex acquire (`handler.rs:187`) is correct — webhook payload is stale after wait.
- Fire-and-forget cache invalidation (`handler.rs:363`) is acceptable; api-server cache has 10-min TTL.
- Silent-skip integration tests in `handler_tests.rs:50` consistent with stated strategy.

### Verdict

**BLOCK** until the two P1 fixes land. Happy path works, but SCA recovery is unmergeable as shipped: pending PI id may be unusable, the system keeps charging while already in pending-action, and recovery links self-destruct on transient 503s.

Pagination + leftover test route can go in the same cleanup commit.

— Reviewed with codex CLI, `model_reasoning_effort="high"`, ~418k tokens.

lit-api-server/src/internal/routes.rs
```diff
@@ -0,0 +1,374 @@
...
+
+fn to_response(identity: ResolvedIdentity) -> Json<VerifyWalletAuthResponse> {
+    Json(VerifyWalletAuthResponse {
+        wallet_address_hex: identity.wallet_address_hex,
+        api_key_hash_hex: identity.api_key_hash_hex,
+    })
+}
+
+#[post("/internal/verify_wallet_auth", format = "json", data = "<body>")]
+pub async fn verify_wallet_auth(
```
glitch003: why are we adding this endpoint in the TEE / lit-api-server?  can't we just import same rust functions into lit-payments and avoid hitting the TEE for this?

lit-api-server/src/internal/routes.rs
```diff
@@ -0,0 +1,374 @@
...
+    };
+    resolver
+        .verify_wallet_auth(&payload)
+        .await
+        .map(to_response)
+        .map_err(map_resolver_err)
+}
+
+#[post("/internal/resolve_api_key", format = "json", data = "<body>")]
+pub async fn resolve_api_key(
```
glitch003: same question as above re: this endpoint.  why not just import the function into lit-payments?

lit-billing-auth/src/lib.rs
glitch003, line 1: do we need a new lit-billing-auth crate?  i would think that auth is just part of lit-billing-core.

lit-payments/src/auto_topup/webhook/mutex.rs
```diff
@@ -0,0 +1,54 @@
...
+use moka::sync::Cache;
+use tokio::sync::Mutex;
+
+const MUTEX_CACHE_CAPACITY: u64 = 10_000;
+const MUTEX_CACHE_TTL: Duration = Duration::from_secs(300);
+
+/// Shared mutex registry. Build once at startup and hand to Rocket state.
+#[derive(Clone)]
+pub struct PerCustomerMutex {
+    inner: Cache<String, Arc<Mutex<()>>>,
```
glitch003, line 30: Why did you choose a moka cache to store the mutexes?

lit-payments/src/auth_resolver.rs
```diff
@@ -0,0 +1,114 @@
...
+impl AuthResolver for HttpAuthResolver {
+    async fn verify_wallet_auth(
+        &self,
+        payload: &WalletAuthPayload,
+    ) -> Result<ResolvedIdentity, AuthError> {
+        let body = VerifyWalletAuthRequest {
+            typed_data: &payload.typed_data,
+            signature: &payload.signature,
+        };
+        let r = self.post("/internal/verify_wallet_auth", &body).await?;
```
glitch003: yeah like right here i would just import verify_wallet_auth.  i don't understand why we would hit the lit-api-server for this

glitch003 (dismissed):
overall it looks good - i think there's some minor changes needed

DashKash54:
## Codex per-phase review — PR #448 (auto-topup, 8 phases + post-glitch fixes)

Independent senior-Rust review per phase against the plan's §17 phase goals and the §6/§9/§11/§12 correctness/concurrency invariants. Findings ranked P0 (blocks merge) / P1 (must fix before prod) / P2 (should fix soon) / P3 (nice-to-have). Each phase reviewed against its own diff vs base `f34705f8`; post-glitch reviewed `607c84d7..7ac53be5`.

---

### Phase 1 — Internal-auth scaffolding + migration (`f34705f8..098e1c15`)

Severity: 0 P0 / 2 P1 / 2 P2 / 1 P3

Top findings:
- **P1** `lit-payments/migrations/20260605000001_auto_topup.sql:26` — `disabled_reason` is free-form `TEXT`. Plan constrains it to `NULL | manual | failures | card_invalid | requires_action`; without a CHECK, later logic can silently persist invalid states and break dashboard/reconciler branching.
- **P1** `lit-api-server/src/internal/guard.rs:39` — rejects length mismatches before `ct_eq`, violating §15's constant-time compare requirement (leaks expected secret length). Use a length-folding constant-time verifier.
- **P2** `lit-api-server/src/internal/config.rs:28` — treats missing `LIT_INTERNAL_SHARED_SECRET` as `None`; allowing prod to boot with internal endpoint unauthorized is a misconfiguration trap. Fail startup outside explicit local/test mode.

---

### Phase 2 — `lit-billing-auth` shared crate + resolver (`098e1c15..2cd3c00d`)

Severity: 0 P0 / 3 P1 / 1 P2 / 0 P3

Top findings:
- **P1** `lit-billing-auth/src/guard.rs:55` — `LIT_DEV_WALLET_BYPASS` accepts arbitrary `X-Dev-Wallet` as `WalletSigned`. Shared between both services, so a bad env rollout becomes full billing auth bypass. Compile-gate behind a test/dev-only feature unavailable in release deploys.
- **P1** `lit-api-server/src/auth_resolver.rs:88` — API-key resolver derives `api_key_hash_hex` from the resolved *wallet bytes* instead of the raw API key. Any lit-payments cache/customer lookup using this hash will key incorrectly.
- **P1** `lit-api-server/src/auth_resolver.rs:70` — Error classification string-matches anyhow text (`"no wallet address"` / `"Address::ZERO"`). Wording drift turns bad keys back into 503 retry loops. Return a typed error from `get_billing_wallet_address`.

---

### Phase 3 — `POST /billing/setup_intent` (`2cd3c00d..e8772ea9`)

Severity: 0 P0 / 1 P1 / 2 P2 / 1 P3

Top findings:
- **P1** `lit-billing-auth/src/guard.rs:76` — `LIT_DEV_WALLET_BYPASS` lets any caller impersonate an arbitrary wallet on `POST /billing/setup_intent`, so a prod env mistake enables attaching a PaymentMethod to another wallet's Stripe customer. Must not live in the shared production guard.
- **P2** `lit-payments/src/billing/setup_intent.rs:48` — API-key auth is accepted by guard but handler returns 501. Breaks valid API-key callers and contradicts the "behind shared auth" goal.
- **P2** `lit-payments/src/billing/setup_intent.rs:95` — idempotency key is a random per-request UUID; client/network retries create duplicate SetupIntents. Use a stable client-supplied key.

---

### Phase 4 — `auto_topup_config` CRUD + DB (`e8772ea9..1a933320`)

Severity: 0 P0 / 2 P1 / 2 P2 / 0 P3

Top findings:
- **P1** `lit-payments/src/billing/auto_topup_config.rs:51` — `resolve_caller` rejects API-key callers with 501; Phase 4 goal is config CRUD behind shared auth and the guard already verifies API keys. Blocks API-key users from reading/disabling config.
- **P1** `lit-payments/src/auto_topup/db.rs:116` — `enabled=true` UPSERT preserves `pending_action_pi_id`/`recovery_token` unconditionally even when `payment_method_id` or top-up params change. Leaves a stale pending SCA PI for an old card/amount.
- **P2** `lit-payments/src/billing/auto_topup_config.rs:108` — PM-ownership validation lists only the first 100 card PMs without pagination; legitimate customers with >100 saved cards get false rejections.

---

### Phase 5 — `customer.updated` webhook trigger + PI create (`1a933320..573977f5`)

Severity: 0 P0 / 2 P1 / 3 P2 / 0 P3

Top findings:
- **P1** `lit-payments/src/auto_topup/webhook/handler.rs:319` — `paymentIntents.create` network/timeout errors return 503; Stripe webhook retry can create a second PI after the 24h Stripe idempotency TTL. Plan §6 says ack 200 and let reconciler recover orphaned succeeded PIs.
- **P1** `lit-payments/src/auto_topup/webhook/handler.rs:462` — after inserting `auto_topup_credits`, a `balance_transactions` timeout returns 503; webhook replay hits `ON CONFLICT` and returns without retrying the NULL `stripe_balance_transaction_id`. Phase 5 has no reconciler yet → credit can remain unapplied until Phase 6.
- **P2** `lit-payments/src/auto_topup/webhook/handler.rs:514` — `month_spend_cents` treats `requires_action`/pending PIs as non-failed spend; abandoned SCA attempts can consume the monthly cap before any credit is delivered.

---

### Phase 6 — Reconciler tokio task (`573977f5..cfc3ddb5`)

Severity: 0 P0 / 1 P1 / 1 P2 / 0 P3

Top findings:
- **P1** `lit-payments/src/auto_topup/reconciler.rs:85` — Orphaned succeeded PIs with no DB row are only swept for `enabled=true` customers; a user who disables auto-topup after being charged but before the reconciler tick remains paid-but-uncredited forever.
- **P2** `lit-payments/src/auto_topup/reconciler.rs:218` — Recovery actions only emit `tracing::warn`; Phase 6 requires log/alert on created or fixed rows. Production may miss load-bearing recovery activity.

---

### Phase 7 — SCA recovery endpoints + dashboard page (`cfc3ddb5..6e9eb077`)

Severity: 0 P0 / 3 P1 / 1 P2 / 0 P3

Top findings:
- **P1** `lit-payments/src/billing/setup_intent.rs:58` / `auto_topup_config.rs:70` — API-key callers get 501 while `auto_recharge.js:75` sends `X-Api-Key` in API mode. Violates shared-auth/API-key requirement and blocks API-key Stripe users from Phase 7.
- **P1** `lit-payments/src/auto_topup/webhook/handler.rs:570` — SCA email links to `/recover_topup?token=...` but the dashboard page added is `recover_topup.html`; on a static dashboard this 404s and strands `requires_action` users.
- **P1** `lit-static/dapps/dashboard/recover_topup.js:45` — recovery creates a Payment Element and calls `confirmPayment` at `:59`, not the planned `stripe.confirmCardPayment(client_secret)` flow for the existing pending PI. Risks failing the Phase 7 SCA gate.

---

### Phase 8 — Dashboard UI + hardening (`6e9eb077..607c84d7`)

Severity: 0 P0 / 5 P1 / 1 P2 / 0 P3

Top findings:
- **P1** `lit-payments/src/billing/setup_intent.rs:58` / `auto_topup_config.rs:70` — API-key callers still 501; locked plan requires identical wallet-sig and API-key auth for dashboard endpoints before prod.
- **P1** `lit-payments/src/auto_topup/webhook/handler.rs:242` — Auto-disable-after-3-failures path only logs; Phase 8 requires the auto-disabled email template/notification.
- **P1** `lit-payments/src/auto_topup/webhook/handler.rs:354` — Declined/expired/insufficient-funds charges only log; Phase 8 requires the declined-payment email with reason and update-card link.
- **P1** `lit-payments/src/auto_topup/webhook/handler.rs:570` — Same SCA-email-vs-`recover_topup.html` 404 mismatch persists.
- **P1** `lit-payments/src/auto_topup/reconciler.rs:84` — Orphan-PI reconciliation still only scans enabled customers; succeeded PI with no DB row remains uncredited if user disables auto-topup before next sweep.
- **P2** `lit-payments/Cargo.toml:8` — No metrics dependency/instrumentation added for trigger count, charge success/failure rate, reconciler activity, or mutex contention. Explicitly part of Phase 8 hardening gate.

---

### Post-glitch fixes (`607c84d7..7ac53be5`)

Severity: 0 P0 / 2 P1 / 1 P2 / 0 P3

Top findings:
- **P1** `lit-payments/src/auto_topup/reconciler.rs:170` — Partial-credit repair retries arbitrarily old rows with `credit:{pi_id}`; after Stripe's idempotency-key window expires, a prior succeeded balance transaction (DB-marking failed) can be credited a second time, violating §11/§12.
- **P1** `lit-payments/src/auto_topup/reconciler.rs:131` — DB-driven repair trusts `customer_id` and `amount_cents` from `auto_topup_credits` but only verifies PI status/source; should also require Stripe PI `customer` and `amount` to match before issuing a balance transaction.
- **P2** `lit-payments/src/auto_topup/webhook/handler.rs:594` — SCA email-send failure is logged but the webhook still returns success after staging `pending_action_pi_id`, leaving the user with no recovery link and future triggers skipped.

---

## Verdict

**0 P0 / 21 P1 / 14 P2 / 1 P3 across the entire PR.**

No merge-blocking issues found by codex. The dominant P1 themes are:

1. **`LIT_DEV_WALLET_BYPASS` in the shared prod guard** (Phases 2, 3) — should be compile-gated, not env-gated, since it grants wallet impersonation.
2. **API-key callers receive 501** on billing/setup endpoints (Phases 4, 5, 7, 8) — the guard accepts API keys but handlers don't honor the resolved identity, breaking Phase 2's shared-auth contract.
3. **SCA recovery URL mismatch** (`/recover_topup` vs `recover_topup.html`, Phases 7 & 8) — 404s on static dashboard.
4. **Reconciler scope** — orphan-PI sweep only covers `enabled=true` customers (Phases 6 & 8) and partial-credit repair lacks PI-amount/customer cross-verification (post-glitch) — both can violate §11/§12 exactly-once invariants.
5. **Webhook 503 on Stripe call failure** (Phase 5) — plan §6 calls for ack-200 + reconciler recovery instead of letting Stripe retry into duplicate PIs.
6. **Missing email notifications** for auto-disable and declined-payment paths (Phase 8) and missing metrics/alerting (Phase 8) — hardening-gate items.

Recommend addressing the `LIT_DEV_WALLET_BYPASS` and API-key-501 items before prod even if technically non-blocking, as they sit on the shared auth boundary (§15).

DashKash54:
## Codex final review — PR #448 (post 11-P1-fix round, pre-merge gate)

Independent Codex CLI review (v0.130.0) across all 8 phases plus the 11-item P1-fix delta (`607c84d7..b9e164a7`). Diffs computed locally against base `f34705f8`. Each phase reviewed against the canonical plan's §6/§9/§11/§12 invariants, §15 service-auth requirements, and §13 accepted trade-offs.

Note: Codex evaluated several phases against the **current working tree** (post-glitch-refactor), not just the pasted diff snapshot, and explicitly says some prior P1s "are already fixed in the checkout."

---

### Phase 1 — Migration, env vars, X-Internal-Secret guard, cache-invalidation endpoint
**Verdict: LGTM. 0 P0 / 0 P1 / 0 P2 / 0 P3.**

Migration matches planned schema/constraints, cache-invalidation endpoint guarded, secret not logged/echoed, degraded-path semantics intentional.

Sub-flag (not blocking): `lit-api-server/src/internal/guard.rs:39` rejects length mismatches before `ct_eq`, which technically leaks expected secret length. Not a merge blocker given fixed high-entropy secret + uniform mismatch responses.

---

### Phase 2 — lit-billing-auth crate + AuthResolver trait
**Verdict: LGTM. 0 P0 / 0 P1 / 0 P2 / 0 P3.**

Verified shared `BillingAuth` guard, EIP-712 verifier, on-chain resolver, both service resolver wirings, and `X-Internal-Secret` guard. Codex confirmed in checkout: dev wallet bypass is debug-only, auth resolution is in-process on both services, extra internal HTTP auth endpoints are gone, API-key auth verifies on-chain with 401/503 split, precomputed hashes rejected in API-key headers, internal-secret compare is constant-time hardened.

---

### Phase 3 — POST /billing/setup_intent
**Verdict: LGTM. 0 P0 / 0 P1 / 0 P2 / 0 P3.**

API-key callers now supported in `setup_intent`, internal-secret guard uses padded constant-time comparison, dev wallet bypass compiled behind `debug_assertions`. (Codex could not run `cargo check` due to sandbox.)

---

### Phase 4 — GET/PUT /billing/auto_topup_config + ownership
**Verdict: 0 P0 / 2 P1 / 0 P2 / 1 P3.**

Codex flagged (against the phase-end snapshot):

- **P1** `lit-billing-auth/src/guard.rs:61` — `LIT_DEV_WALLET_BYPASS` production-code auth bypass. *Status: addressed in current checkout per Phase 2's independent verification ("dev wallet bypass is debug-only").*
- **P1** `lit-payments/src/billing/auto_topup_config.rs:60` — API-key callers get 501 from GET/PUT config. *Status: addressed by P1-fix item #1 in the delta commit (`b9e164a7`), confirmed via Phase 3 LGTM ("API-key callers are now supported").*
- **P3** `auto_topup_config.rs:112` — payment-method ownership only reads first `limit=100` page. *Explicitly deferred per PR scope (pagination >100 cards = theoretical).*

Effective post-delta: **0 P1** outstanding.

---

### Phase 5 — customer.updated webhook + sync charge + sync credit
**Verdict: 0 P0 / 2 P1 / 0 P2 / 0 P3.**

- **P1** `lit-payments/src/auto_topup/webhook/handler.rs:328` / `:359` — Stripe 429 / rate-limit / lock-timeout responses that are not HTTP 5xx fall through to `return Ok(())`. If no PI was created, reconciler cannot recover. Violates §7's "5xx on transient Stripe API errors" and bias-toward-more-topups invariant. **Fix: classify Stripe 429/transient error types as `ProcessError::Transient`.**
- **P1** `lit-payments/src/auto_topup/reconciler.rs:160` — partial credit rows older than 23h are skipped indefinitely / manual-only. If idempotency key expired, needs automated verification path before skipping. (Duplicated in Phase 6 + 8 findings.)

---

### Phase 6 — Reconciler tokio task
**Verdict: 0 P0 / 2 P1 / 0 P2 / 0 P3.**

- **P1** `lit-payments/src/auto_topup/reconciler.rs:101` — **Disabled-with-orphan gap still open.** Pass B scan is `enabled ∪ recently_active` where "recently active" reads from `auto_topup_credits`. If `paymentIntents.create` succeeds, process crashes before `try_insert_credit`, then manual-ops disables config: no credit row + `enabled=false` → charged PI never credited. Violates §9/§12 "HTTP timeout on PI create → reconciler credits." **Fix: scan ALL `auto_topup_config` customers in the lookback window, or persist an attempted-charge marker before PI create.**
- **P1** `lit-payments/src/auto_topup/reconciler.rs:160` — same 23h partial-skip issue as Phase 5. Violates §9's 7-day recovery window. **Fix: reconcile against existing customer balance transactions by PI description/amount before skipping; if none, apply deliberate one-time repair path.**

Constant-time webhook/internal-secret comparison: clean.

---

### Phase 7 — SCA resume endpoints
**Verdict: LGTM. 0 P0 / 0 P1 / 0 P2 / 0 P3.**

SCA endpoints preserve token on transient Stripe failure, consume on successful client-secret retrieval, verify `succeeded` + `metadata.source=auto_topup` before crediting, reuse `auto_topup_credits` unique row + `credit:{pi_id}` idempotency key, repair partial credit rows immediately, clear pending state after credit. P1 fix in `post_auto_topup_resume_complete` matches plan's sync-credit and reconciler invariants.

---

### Phase 8 — Dashboard UI + SCA recovery page + CORS
**Verdict: 0 P0 / 1 P1 / 0 P2 / 0 P3.**

- **P1** `lit-payments/src/auto_topup/reconciler.rs:154` + `lit-payments/src/billing/sca_resume.rs:249` — Same 23h partial-skip issue (duplicate of Phase 5/6 finding). Recommends reconciling old partials by checking Stripe customer balance transactions for existing `Auto top-up via {pi_id}` credit before writing, or deterministic repair path that still credits within the 7-day sweep.

Constant-time webhook/internal-secret checks + CORS hardening: clean.

---

### Codex-P1-fix delta review (607c84d7..b9e164a7 — the 11-item follow-up commit)
**Verdict: 0 P0 / 1 P1 / 0 P2 / 0 P3.**

10 of 11 items pass cleanly. One item incomplete:

- **P1** `lit-payments/src/auto_topup/db.rs:348` / `lit-payments/src/auto_topup/reconciler.rs:104` (Item #7 — reconciler scope union) — **The disabled-with-orphan gap is still open.** `list_recently_active_customers()` defines "recently active" via recent `auto_topup_credits` rows, but the orphan case this is meant to fix is specifically "PI succeeded, service crashed before INSERT" — in which case there is no `auto_topup_credits` row. If config is later `enabled=false`, `run_once()` will not scan that customer's Stripe PIs. Existing test only covers an enabled no-row orphan, missing the regression. **Fix: use a source that exists before the crash point — recent `auto_topup_config.updated_at` rows, all configured customers in lookback window, or persist an attempted-charge marker before PI create.**

Items 1, 2, 3, 4, 5, 6, 8, 9, 10, 11: no regressions flagged.

---

## Final verdict

**0 P0 / 3 unique P1 / 0 P2 / 1 P3** across the whole PR (de-duplicating overlapping findings).

The 3 unique P1s:
1. **Disabled-with-orphan reconciler gap** (Phase 6, Phase 8, P1-fix-delta) — `auto_topup_credits`-based "recently active" cannot recover a crash *before* the credit row exists when config is then disabled. Most severe — directly violates §9/§12 load-bearing invariant.
2. **Stripe 429 / non-5xx transient classification** (Phase 5) — currently swallows as `Ok(())`, no reconciler recovery for top-ups where no PI was created.
3. **23h partial-credit skip** (Phase 5, Phase 6, Phase 8) — partial rows past Stripe's idempotency window are permanently skipped, violating §9's 7-day recovery window without an automated cross-verify path.

**RECOMMENDATION: FIX BEFORE MERGE.** All three P1s touch the load-bearing reconciler / webhook-recovery path explicitly called out in the plan as the safety net for "charged-but-uncredited" users. Each has a concrete suggested fix. Phases 1/2/3/7 are independently LGTM'd.

lit-api-server/src/internal/routes.rs
DashKash54: Moved the auth core into a shared crate that both services import directly

lit-billing-auth/src/lib.rs
DashKash54: Folded into lit-billing-core as a billing_auth module

DashKash54:
Not addressing the final review, we have a recon that checks every 15 mins, we have manual intervention by Sally, this is a edge case which we have mostly handled

lit-payments/src/auto_topup/webhook/mutex.rs
DashKash54: Moka was already a direct dep (auth/rate_limit.rs). TTL/bounded eviction came along for free — DashMap would also have worked, but since it was already there I reused it

DashKash54:
## Codex final pre-merge review (post browser-QA fixes)

No P0 merge blockers found.

**Findings**
- P1 `lit-payments/src/auto_topup/webhook/handler.rs:212` — stale-pending escape hatch clears live SCA recoveries whose Stripe PI is `requires_payment_method`.
  The recovery page explicitly handles the post-`authentication_required` shape where the PI is `requires_payment_method` and re-confirms with `confirmCardPayment(..., { payment_method })`. The escape hatch treats any status other than `requires_action` as stale, calls `clear_pending_action_force`, invalidates the token, and may create a fresh PI/email on the same webhook. Those old SCA PIs also count as failures at `handler.rs:591`, so repeated balance changes can incorrectly trip auto-disable. Treat `requires_payment_method` + `last_payment_error.code=authentication_required` as recoverable until token expiry, or require both token expiry and stale Stripe status before clearing.

- P2 `lit-billing-core/src/on_chain.rs:270` — selector helper/test still returns/pins `getBillingWalletAddress(bytes32)`.
  The production calldata path at `on_chain.rs:144` uses `getBillingWalletAddress(uint256)` and appends the 32-byte hash directly, which matches the stated Solidity ABI and is correct. But the exported helper and pinned test still use `bytes32`, so any mock/downstream caller using the helper will test the wrong selector.

- P3 `lit-payments/src/auto_topup/db.rs:460` — recovery token exchange is no longer strictly single-use under concurrent GETs.
  `lookup_recovery_token` does a plain SELECT, then `sca_resume.rs:136` clears after Stripe succeeds. Two simultaneous GETs can both receive `client_secret` and `payment_method_id`. Impact is low because both target the same PI and crediting is deduped, but it is weaker than the plan's one-time bearer-token model.

**Notes**
- `GET /billing/payment_method` is auth-guarded and ownership-checked; no meaningful new cross-tenant exposure there.
- The three credit-concurrency layers still hold: per-customer mutex, `auto_topup_credits.payment_intent_id` uniqueness, and `credit:{pi_id}` Stripe idempotency key.
- Source review only; tests were not run.

---
Generated with Codex CLI via Claude Code

GTC6244:
Verified clean (the high-risk stuff)
- Stripe webhook signature: v1 HMAC-SHA256 over t.payload, constant-time compare, ±300s skew, verified before JSON parse. ✓
- eip712.rs move (1088 lines): faithful relocation to lit-billing-core, only mechanical import/error-type changes + chain_id now a param. No silent behavior change. ✓
- Authorization: every config/setup endpoint derives customer_id from the verified caller’s wallet, never the request body; cross-tenant PM theft blocked and tested. ✓
- CORS: lit-payments correctly moved from AllowedOrigins::all() to an exact-match allowlist (closes the credentials+wildcard CSRF combo). ✓
- No XSS: all dynamic values go through textContent; the only innerHTML writes are constant clears. ✓
- Migrations: additive, safe on the brand-new tables; final constraint state is coherent (topup $5–$200, optional monthly cap ≥ topup). ✓
- cargo fmt --check clean on all three crates; lit-billing-core builds + 61 tests pass; no k6 regen needed (new internal route isn’t #[openapi]).

PR #449, opened 2026-06-08
feat(auto-topup): phase 1 — schema + env vars + cache-invalidation endpoint

DashKash54:
**Stack:** PR 1 of 8. Base: `DashKash54/stripe-metadata-vs-postgres`. Next: [#450](https://github.com/LIT-Protocol/chipotle/pull/450).

## What this PR does

Ships the foundation layer for the Stripe auto top-up feature. Three pieces, none of which contain auto-topup business logic yet — but every subsequent phase depends on them.

1. **Postgres schema** for the two new tables auto top-up needs.
2. **Env vars + shared-secret guard** so lit-payments can call lit-api-server (and vice versa) over an authenticated internal channel.
3. **`POST /internal/invalidate_balance_cache`** on lit-api-server — the only inbound internal call lit-api-server accepts from lit-payments. Phase 5 fires this after a successful sync credit so the api-server's balance cache drops the stale entry.

## How it's implemented

* `lit-payments/migrations/20260605000001_auto_topup.sql`:
  - `auto_topup_config` (per-user rule). Primary key `customer_id`, `UNIQUE(wallet_address)`, includes SCA-recovery columns (`pending_action_pi_id`, `pending_action_at`, `recovery_token`, `recovery_token_expires_at`) used by Phases 5 + 7. `CHECK enabled_requires_config` enforces "`enabled=true` ⇒ threshold/topup/cap/pm/consent all set".
  - `auto_topup_credits` (per-PI dedup ledger). PK on `payment_intent_id` so the Phase 5 webhook handler + Phase 6 reconciler converge on exactly one credit per PI.
  - Two partial indexes: `recovery_token IS NOT NULL` (Phase 7 token lookup) and `stripe_balance_transaction_id IS NULL` (Phase 6 reconciler's "partial credit" scan).
* `lit-payments/src/config.rs`: adds `LIT_API_SERVER_BASE_URL`, `LIT_INTERNAL_SHARED_SECRET`, `STRIPE_WEBHOOK_SECRET`, optional `RECONCILER_INTERVAL_SECS` (default 900) to `Config::from_env()`.
* `lit-payments/src/internal/client.rs`: reusable `reqwest::Client` builder + `post_internal()` helper for outbound calls to lit-api-server. Carries the `X-Internal-Secret` header.
* `lit-api-server/src/internal/{config,guard,routes}.rs`:
  - `InternalConfig::init()` reads `LIT_INTERNAL_SHARED_SECRET`, returns `None` when absent (mirrors `stripe::init()` pattern so dev/simulator builds don't crash without it).
  - `InternalSecret` Rocket request guard, constant-time compare via `subtle::ConstantTimeEq`. 401 on missing/wrong/length-mismatched header.
  - `POST /internal/invalidate_balance_cache` route, body `{customer_id}`, calls a new public method `StripeState::invalidate_balance_cache(&self, customer_id: &str)` that wraps the existing private `balance_cache.invalidate()` (mirrors precedent at `stripe.rs:613`).

## How it's been tested

* `cargo test --lib internal::` in `lit-api-server`: 4 Rocket integration tests using `rocket::local::asynchronous::Client` (same pattern as `health.rs`):
  - `accepts_correct_secret` → 200
  - `rejects_missing_header` → 401
  - `rejects_wrong_secret` → 401
  - `rejects_secret_of_different_length` → 401
* `cargo check` clean on both crates.
* `psql \d` against a freshly-migrated local Postgres confirms both tables, all columns, primary keys, the unique constraint, both partial indexes, and the `enabled_requires_config` CHECK.
* `INSERT … enabled=true` with null fields verified live to be rejected by the CHECK constraint.

## Key areas for review

* **Schema correctness** — `lit-payments/migrations/20260605000001_auto_topup.sql`. CHECK predicate, partial-index `WHERE` clauses, and SCA columns are load-bearing for later phases.
* **Constant-time compare** — `lit-api-server/src/internal/guard.rs:50–55`. Length-shortcircuit is intentional (length of expected secret isn't itself secret); the byte comparison uses `subtle::ConstantTimeEq`.
* **Cache-invalidation public surface** — `lit-api-server/src/stripe.rs` adds `impl StripeState { pub async fn invalidate_balance_cache(&self, ...) }`. Verify it's exactly what Phase 5's webhook handler will call.
* **Env-var failure mode** — `InternalConfig::init()` returns `None` when missing, which means routes that depend on it fail closed. Confirm that's the right shape for the TEE deploy.

### DIFF
--- a/lit-api-server/Cargo.lock
+++ b/lit-api-server/Cargo.lock
@@ -4786,6 +4786,7 @@ dependencies = [
  "sqlx",
  "strum 0.28.0",
  "strum_macros 0.28.0",
+ "subtle",
  "tdx-quote",
  "tokio",
  "toml_edit 0.22.27",
--- a/lit-api-server/Cargo.toml
+++ b/lit-api-server/Cargo.toml
@@ -76,6 +76,7 @@ rand = { version = "0.8" }
 sqlx = { version = "0.8", features = ["sqlite", "runtime-tokio"] }
 strum = "0.28.0"
 strum_macros = "0.28.0"
+subtle = "2.6"
 url = "2"
 
 rocket_okapi = { version = "=0.9.0", features = ["swagger"] }
--- a/lit-api-server/src/internal/config.rs
+++ b/lit-api-server/src/internal/config.rs
@@ -1,0 +1,38 @@
+//! Env-driven configuration for the inbound `lit-payments` → `lit-api-server`
+//! channel.
+//!
+//! Initialised once at boot and shoved into Rocket-managed state. Mirrors the
+//! `stripe::init()` pattern: returns `None` when the env var is missing so
+//! that local dev / TEE simulator runs without auto-top-up wiring keep
+//! working.
+//!
+//! Required env var:
+//!   • `LIT_INTERNAL_SHARED_SECRET` — shared with `lit-payments`. Compared in
+//!     constant time on inbound `X-Internal-Secret` headers from
+//!     `lit-payments`'s sync-credit handler. `lit-api-server` does NOT call
+//!     `lit-payments` in the new design (the trigger is the `customer.updated`
+//!     webhook), so no outbound URL is needed here.
+
+use std::sync::Arc;
+
+#[derive(Clone, Debug)]
+pub struct InternalConfig {
+    /// Shared secret used by the [`super::guard::InternalSecret`] request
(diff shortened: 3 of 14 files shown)


DashKash54:
Ported to PR 448

PR #453, opened 2026-06-08
feat(auto-topup): phase 5 — customer.updated webhook + sync charge + sync credit

DashKash54:
**Stack:** PR 5 of 8. Base: `auto-topup-phase-4`. Diff shows only Phase 5.

## What
Load-bearing trigger phase. Implements the full §6 flow synchronously inside the webhook request: HMAC verification → quick exits → per-customer mutex → fresh balance fetch → PI list / failure / cap → off-session PaymentIntent create → INSERT auto_topup_credits ON CONFLICT DO NOTHING → balance_transactions write with credit:{pi.id} idempotency key → UPDATE row with bt_id → fire-and-forget cache invalidation.

* Raw `Data` handler (NOT Json — HMAC verifies exact bytes), 1 MiB body cap.
* HMAC-SHA256 signature verification, ±300s timestamp skew, multi-v1 rotation.
* Per-customer moka mutex (5-min TTL) — optimization only; correctness from PI ledger UNIQUE + Stripe idempotency keys.
* Branches on PaymentIntent response: `succeeded` → sync credit; `authentication_required` → save `pending_action_pi_id` + recovery_token (Phase 7 consumer); decline codes → log + 200; other → 503 retry.

**Codex P1 #1 + #2 fixes folded in:**
* StripeError struct in lit-billing-core preserves `error.payment_intent.id` so SCA recovery can stage the right PI id (not a regex'd error string).
* Webhook short-circuit gates on `enabled && pending_action_pi_id.is_none()` so an in-flight SCA can't be overwritten by a fresh `customer.updated`.

## Tests
* 19 unit tests (signature verification, SCA token gen, handler helpers).
* 9 integration tests against real Stripe + real Postgres: tampered/stale signature, non-customer.updated event, no-balance event, disabled, above-threshold, happy path, replay safety, cap reached, **pending_action_pauses_further_topups** (codex P1 #2 regression).

### DIFF
--- a/lit-billing-core/src/http.rs
+++ b/lit-billing-core/src/http.rs
@@ -1,5 +1,7 @@
 //! Pure HTTP response parsing for the Stripe API.
 
+use std::fmt;
+
 use anyhow::Result;
 use reqwest::StatusCode;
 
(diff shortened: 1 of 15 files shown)


DashKash54:

### Fix-specific review (`573977f5`)

No P0/P1 issues found in this diff.

Reviewed the introduced changes around config re-read under mutex, event-id-derived PI idempotency key, and Stripe 5xx handling in `lit-payments/src/auto_topup/webhook/handler.rs`; no merge-blocking or must-fix-before-prod regression.

### Whole-PR re-review (`auto-topup-phase-4..auto-topup-phase-5`)

No P0s found.

**P1**
- `lit-payments/src/auto_topup/webhook/handler.rs:401`: partial credit retries are swallowed. If `try_insert_credit` succeeds but the balance transaction or `mark_credit_completed` fails, the webhook returns 503. On retry, the same PI hits `ON CONFLICT`, `claimed == false`, and the handler returns 200 without checking whether `stripe_balance_transaction_id` is still `NULL`. That can leave a successful charge uncredited and stop Stripe retries. The conflict path needs to resume/verify completion or keep returning transient while incomplete.

- `lit-payments/src/auto_topup/webhook/handler.rs:196` / `lit-payments/src/auto_topup/db.rs:215`: the duplicate-charge/SCA guard is still only in-process and `set_pending_action` is a blind update. In a multi-replica deployment, two different `customer.updated` events for the same customer can both pass the low-balance/cap checks, create different PIs, and a second SCA path can overwrite the first recovery token. This needs a durable per-customer claim/state transition, row lock, advisory lock, or conditional `pending_action_pi_id IS NULL` update with rows-affected handling.

**Prior P1 Status**
- Structured SCA PI extraction appears addressed: `StripeError` preserves the JSON body and handler reads `error.payment_intent.id` via downcast.
- Stable idempotency for retries of the same Stripe event appears addressed via `auto_topup_pi:{event_id}`.
- Post-mutex config reload / pending-action short-circuit is addressed for same-process concurrency, but not across replicas as noted above.
- Structured Stripe 5xx from `paymentIntents.create` now returns transient/503, so that prior issue appears addressed.

---

DashKash54:
Ported to PR 448

PR #455, opened 2026-06-08
feat(auto-topup): phase 7 — SCA resume endpoints with codex P2 token-preserve fix

DashKash54:
**Stack:** PR 7 of 8. Base: `auto-topup-phase-6`. Diff shows only Phase 7.

## What
Plan §6a SCA recovery flow. Two unauthenticated endpoints driven by the one-time `recovery_token` issued by the Phase 5 webhook handler when an off-session PI returns `authentication_required`.

* **GET /billing/auto_topup_resume?token=…** — non-consuming lookup, fetches the PI's client_secret from Stripe, then atomically clears the token only AFTER the Stripe retrieve succeeds. Transient Stripe failures leave the token usable for retry. Single-use guarantee for the happy path is preserved (success burns the token; reload after success → 404). **Codex P2 #3 fix incorporated.**
* **POST /billing/auto_topup_resume/complete** — body `{payment_intent_id}`. Re-fetches the PI server-side, verifies `status=succeeded` AND `metadata.source=auto_topup`, runs the sync credit path. Idempotent — repeated calls don't double-credit.

DB primitives:
* `lookup_recovery_token` (read-only)
* `clear_recovery_token_for_pi` (atomic NULL'ing keyed by pi_id)
* `consume_recovery_token` (legacy one-shot, kept dead-allowed)

## Tests
6 integration tests against real Stripe + real Postgres:
* get_resume_with_valid_token_returns_client_secret (+ second-read 404)
* get_resume_with_expired_token_returns_404
* get_resume_with_unknown_token_returns_404
* **get_resume_preserves_token_on_stripe_failure** (codex P2 #3 regression)
* complete_with_succeeded_pi_credits_and_clears_pending
* complete_rejects_non_succeeded_pi

### DIFF
--- a/lit-payments/src/billing/mod.rs
+++ b/lit-payments/src/billing/mod.rs
@@ -7,9 +7,12 @@
 //! `X-Internal-Secret` channel (see `auth_resolver::HttpAuthResolver`).
 
 pub mod auto_topup_config;
+pub mod sca_resume;
 pub mod setup_intent;
 
 #[cfg(test)]
 mod auto_topup_config_tests;
+#[cfg(test)]
+mod sca_resume_tests;
 #[cfg(test)]
 pub(crate) mod setup_intent_tests;
--- a/lit-payments/src/billing/sca_resume.rs
+++ b/lit-payments/src/billing/sca_resume.rs
@@ -235,7 +235,30 @@ pub async fn post_auto_topup_resume_complete(
                 "Could not record your credit; try again.",
             )
         })?;
-    if claimed {
+    // Codex P1 (Phase 7): when claimed=false, a prior path already wrote
+    // the credit row — but it may be a partial (balance_tx_id NULL) left
+    // over from a mid-flow crash on the original webhook delivery. Treat
+    // partials the same as a fresh claim: run the balance_transactions
+    // write under the same idempotency key. The Stripe Idempotency-Key
+    // guarantees we don't double-credit if the prior attempt actually
+    // landed at Stripe but failed on the DB write. Pre-fix this branch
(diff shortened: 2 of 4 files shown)


DashKash54:

### Fix-specific review (`fb2dca1e`)

P0: none found.

P1:
- `lit-payments/src/billing/sca_resume.rs:247` - Treats any existing `NULL stripe_balance_transaction_id` row as replay-safe forever; after Stripe idempotency keys expire, reusing `credit:{pi_id}` can create a second balance transaction and double-credit.
- `lit-payments/src/billing/sca_resume.rs:250` - No fresh-partial age gate, so `/complete` can race the live webhook's in-flight balance write and return a user-visible Stripe/idempotency error instead of the previous OK path.

Stripe basis: https://docs.stripe.com/api/idempotent_requests

### Whole-PR re-review (`auto-topup-phase-6..auto-topup-phase-7`)

**P0/P1 Findings**

None. LGTM for P0/P1.

**Prior P1 Status**

Addressed: partial credit rows are now repaired in `/complete`. `lit-payments/src/billing/sca_resume.rs:247` re-checks an existing credit row and retries the Stripe balance transaction when `stripe_balance_transaction_id` is `NULL`, using the same `credit:{pi_id}` idempotency key at `lit-payments/src/billing/sca_resume.rs:262`.

Addressed: token preservation on Stripe retrieve failure also appears fixed. `lit-payments/src/billing/sca_resume.rs:83` does a non-mutating lookup, and the token is only cleared after successful Stripe retrieval at `lit-payments/src/billing/sca_resume.rs:129`.

---

DashKash54:
Ported to PR 448

PR #475, opened 2026-06-09
ci: detect global Foundry install to avoid reinstall collisions

clawdbot-glitch003:
The `forge-test` and `contract-bindings-check` jobs were reinstalling Foundry on every run even though the shared runner already has it globally in `~/.foundry/bin` — that directory is only on PATH via the user's shell profile, which GH Actions' non-login step shells don't source, so `command -v forge` came up empty and the guard always set `installed=false`. The needless reinstall intermittently collided with another job running `forge` on the shared runner, producing the `'forge' is currently running` error. Both workflows now probe `~/.foundry/bin` explicitly and, when the pinned version is present, export its directory to `$GITHUB_PATH` so later steps reuse it without reinstalling.

### DIFF
--- a/.github/workflows/contract-bindings-check.yml
+++ b/.github/workflows/contract-bindings-check.yml
@@ -34,8 +34,17 @@ jobs:
       - name: Check Foundry toolchain
         id: foundry-toolchain
         run: |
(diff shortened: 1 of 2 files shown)


glitch003:
confirmed that this fixes the CI issue - will merge if all tests pass

PR #483, opened 2026-06-11
feat(lit-venues): native venue connectors + in-TEE proxied fetch op

clawdbot-glitch003:
## What

Adds a **venue layer** that makes centralized exchanges callable from inside Lit Actions with credentials sealed in the TEE, plus the **egress primitive** that makes geo-blocked venues (Binance) reachable from the enclave.

- **`lit-venues/`** — native, REST-only connectors (**Binance** incl. `binance.us` + spot testnet, **Coinbase Advanced Trade**, **Hyperliquid perps**) behind one ccxt-shaped interface. No Node built-ins; `@noble/*` crypto inlined; ~178KB IIFE bundle. Exact decimal-string math; unified `VenueError` taxonomy; HMAC-SHA256 / Ed25519 (Binance), ES256-JWT (Coinbase CDP), and msgpack→EIP-712 (Hyperliquid) signing.
- **`op_lit_proxied_fetch`** (new Lit Actions op, exposed as `Lit.Actions.proxiedFetch`) — egresses in-process via `reqwest` through a per-request authenticated proxy, with per-proxy client pooling. TLS is end-to-end through the proxy CONNECT tunnel; responses capped at 10 MiB; counts against the per-action fetch quota.
- **Email approval primitive (plan D6/M3)** — `Lit.Actions.sendEmail` / `requestEmailApproval` / `checkEmailApproval`, with the attestation verified **in-TEE** (fail closed).
- **Plan**: `plans/ccxt-venue-layer-and-email-approval.md`.

## Update 2026-06-11 — rest of the plan implemented

**Hyperliquid: the first PKP-native venue (D8/M2.5).** No API key exists — every trading action is an EIP-712 signature, and the key is the action-bound TEE key, connected as a venue-side agent wallet that structurally cannot withdraw. Signing (hand-rolled msgpack action hash + phantom agent, chainId 1337; `HyperliquidSignTransaction` user-signed actions) is **pinned byte-for-byte to the official SDK's test vectors**. Perp surface (`fetchPositions`/`setLeverage`/`fetchFundingRate`), `approveAgent`, builder codes, 5-sig-fig/szDecimals rules. Live public conformance 4/4 against the real API — measured: **api.hyperliquid.xyz answers direct US egress, no proxy needed**.

**Email approval (M3), both halves.** Runtime: 3 new ops on the existing op stream; `checkEmailApproval` verifies the `email-approval-v1` attestation inside the runtime against a pinned pubkey — tampered/expired/misbound/unsigned all reject (a compromised approval service can deny, never forge). Server: `ApprovalService` — single-use nonced approvals, L1 link / L2 OTP step-up (OTP returns to the requesting app; email stays a notification channel), k256-signed attestations, per-account daily quotas, strict text-only templates, Resend delivery, `/approvals/<id>` pages + `/approvals_pubkey`. v1 keeps pending approvals in-memory (documented).

**M1/M2 remainder:** markets-cache injection (fetchMarket can cost zero quota), per-proxy reqwest client pooling, `docs/runbooks/venue-proxy-rotation.md`.

**Gates + ops:** `k6/correctness/venues-{binance,coinbase,hyperliquid}.spec.ts` (public always-on; authenticated lifecycles env-gated on testnet/CDP keys), `e2e/tests/api/email-approval.spec.ts` (two-phase L2 incl. wrong-OTP/replay/deny), daily `venue-drift-canary.yml` that files a venue-drift issue on failure.

**M4:** six examples — `venue-portfolio-read`, `funding-rate-monitor`, `venue-twap-order` (D7 chained ticks + markets-cache), `price-trigger-stop` (lit-triggers armed), `cex-sweep-with-email-approval` (D6 two-phase), `hl-agent-perp-policy` (PKP-native showcase).

## Verification

- **lit-venues: 58 unit tests** — Binance docs HMAC vector, RFC 8032 Ed25519, Coinbase ES256-JWT roundtrip, and 15 Hyperliquid signing tests pinned to the official SDK vectors (phantom-agent connectionId, GTC/cloid/TPSL, vault marker, user-signed UsdSend/Withdraw).
- **Live conformance** (`lit-venues/scripts/verify-live.mjs`): Coinbase + Hyperliquid 7/7 direct; Binance tier runs when a proxy is configured.
- **lit-actions: 33 integration tests green** (3 real-network tests `#[ignore]`d as before), including 6 new email-approval tests proving in-TEE rejection of tampered/expired/misbound attestations.
- **lit-api-server: 223 lib tests green**, including 4 ApprovalService tests (L2 OTP flow, single-use, account scoping, quota, expiry).
- **M0 spike previously PASSED on staging** (582ms, US egress confirmed).

## Remaining (environment, not code)

- Venue keys for the authenticated gates: `BINANCE_TESTNET_KEY/SECRET`, `COINBASE_KEY_NAME/PRIVATE_KEY`, `HYPERLIQUID_TESTNET_KEY` (+ canary secrets).
- Dev env config for M3: server `LIT_APPROVAL_ATTESTATION_KEY`, `RESEND_API_KEY`, `LIT_APPROVAL_PUBLIC_BASE_URL`, `LIT_APPROVAL_EXPOSE_LINK=true` (dev only); runtime `LIT_APPROVAL_ATTESTATION_PUBKEY` (pin from `GET /approvals_pubkey`).
- In-TEE proxied-fetch network proof + the e2e approval gate run on the next dev deploy.
- Tiny live orders (Coinbase Advanced Trade, Hyperliquid mainnet) behind manual flags before declaring Tier 1.

## Reviewer notes

- `op_lit_proxied_fetch` uses `#[op2(async(lazy))]` — edition 2024 rejects the bare `async` keyword in the attribute. Proxy auth is applied via `.basic_auth()` because `reqwest::Proxy` doesn't forward URL userinfo. Both documented in-code.
- Hyperliquid msgpack: object-literal key order in `hyperliquid.ts`/`hyperliquid-signing.ts` is load-bearing (the action hash commits to it) — the vector tests will catch any reorder.
- New proto fields 16–18 extend both `ExecuteJs` oneofs; errors ride the existing `ReportError` path.
- `tests/it.rs::proxied_fetch` remains `#[ignore]`d (needs real outbound network): `LIT_VENUES_TEST_PROXY=… cargo test -p lit-actions-tests --test integration proxied_fetch -- --ignored`.


clawdbot-glitch003:
**Live-key conformance: 11/11 PASS** (`verify-live.mjs`, commit e629186b)

- **Coinbase authenticated reads live** — ES256-JWT against a real Advanced Trade account (18 non-zero accounts). Found+fixed: CDP PEMs arrive with literal `\n` from the JSON download; parsers now normalize.
- **Hyperliquid full testnet lifecycle live, with a real agent key** — deep GTC limit rest → list → cancel, plus an IOC fill round-trip: bought 0.01 ETH @ 1739.8, position visible, reduce-only close @ 1738.68, account flat. Run twice.
- Found+fixed a real footgun live testing was for: agent (API-wallet) keys are the default connection mode and their own account is always empty — reads silently targeted the agent instead of its master. The connector now auto-resolves the master via the venue's `userRole` lookup (cached; explicit `accountAddress` overrides).
- Security model observed working as designed: the agent key **could not** move USDC spot→perp (`usdClassTransfer` is user-signed, master-only) — exactly the no-funds-movement guarantee D8 claims.

Remaining for Tier-1 declarations: Binance testnet keys, the same suites as Lit Actions on dev (k6 gates, post-deploy), and the tiny live orders behind manual flags.

clawdbot-glitch003:
Correction to the earlier note: the test account is a Hyperliquid **unified account** (newer venue default — the spot→perp transfer button is disabled there because perps margin straight from the unified/spot balance). That's why fills cleared while our perp-only balance read showed 0. `fetchBalances` now merges perp equity + spot/unified USDC (preferring the venue's `tokenToAvailableAfterMaintenance` for `free`) — it reads **998.86 USDC** on the live account and the full lifecycle + fill round-trip is green again, 11/11. No action needed on the account.

clawdbot-glitch003:
Fixed all 9 codex findings in `0f821cc8`.

**P1 — operation binding (the big one):** the attestation now binds a caller-supplied `requestHash` (hash of the exact operation). The runtime verifies in-TEE that it matches the hash the consuming action expects, and refuses an unbound approval where one is expected. A valid "0.1 BTC to X" approval can no longer be replayed for a different movement. Wired through proto → ops → SDK → `ApprovalService`, and the cex-sweep example now recomputes the same canonical hash in both phases (new `sha256Hex` helper).

**P1 — single-use:** approved records are consumed on first check (`Consumed`); replays return `consumed`, not approved.

**P1 — OTP lockout:** ≤5 wrong attempts then the record locks — online brute-force of the 6-digit code is bounded.

**P1 — honest trust boundary:** the module doc now states plainly that in-TEE verification stops a *Flows/transport* forgery but **not** an approval-*service* compromise (it holds the signing key; signer isolation is the tracked follow-up). `from_env` refuses an ephemeral key unless `LIT_APPROVAL_ALLOW_EPHEMERAL_KEY=true`, and the server degrades gracefully (approval ops report "not configured") instead of panicking all action traffic.

**P2:** hyperliquid monotonic nonce (`max(now, last+1)`); proxied-fetch error no longer interpolates the reqwest URL/chain (Binance signs in the query string); quota map swept + documented as per-usage-key; 4 committed `.DS_Store` removed + gitignored.

Tests: lit-venues 62, lit-actions integration 36 (3 net-ignored) incl. new binding match/mismatch/unbound-refusal + consume tests, lit-api-server approvals 7 incl. consume / OTP-lockout / ephemeral-refusal. The e2e M3 gate now exercises operation-binding match+mismatch and replay→consumed.

glitch003:
I'm closing this PR because I have extracted all of the actions into the Flows repo and am doing a new PR below which only includes the TEE lit API server portion of this PR.

https://github.com/LIT-Protocol/chipotle/pull/502

PR #487, opened 2026-06-11
feat(billing): Stripe webhook to debit credits on refunds/disputes (CPL-335)

GTC6244:
Credits were granted on the client-initiated `POST /billing/confirm_payment` but nothing ever debited them when Stripe clawed the money back, so a card chargeback or refund let a customer keep credits they no longer paid for (unbounded loss). This adds a signature-verified `POST /billing/webhook` endpoint that, on `charge.dispute.created` and `charge.refunded`, writes a positive customer balance transaction debiting the clawed-back amount (`payment_intent.payment_failed` is acknowledged without a debit, since no credit was ever granted). Signature verification (HMAC-SHA256 over the raw body, ±5-min replay window, secret rotation), event parsing, and the event→clawback interpretation live as pure, unit-tested helpers in `lit-billing-core`. Idempotency is anchored on the refund/dispute id via the Stripe `Idempotency-Key` plus an in-process processed-events cache, and the route is mounted outside the OpenAPI surface (like `/health`) so it stays out of `openapi.json` and the generated k6 client. Verified with `cargo build`, `clippy`, `fmt --check`, and the new unit + Rocket integration tests; operational follow-up is to register the endpoint in the Stripe Dashboard and set `STRIPE_WEBHOOK_SECRET`.


GTC6244:
Given the delta, and move to remove almost everything billing related from lit-api-server, closing.

PR #489, opened 2026-06-12
fix(cache): short TTL for denied permission results (stale 403 for newly permitted actions)

clawdbot-glitch003:
## Customer report

Executing a newly registered action returned `HTTP 403: "The provided API key is not authorized to execute the specified action (Qmd…/147…)"` for ~1 hour, even though:

- the on-chain `updateGroup` / `addActionToGroup` transactions had mined (`receipt.status == 0x1`),
- two sibling actions in the same group executed fine with the same usage key,
- retrying after ~1 hour succeeded.

## Root cause

The customer's account mutates permissions via **wallet-signed transactions submitted directly on-chain** (ChainSecured mode). The API server never observes those writes, so `blockchain_cache::invalidate_for_account` — which covers all of the server's own mutation endpoints — never fires.

A `canExecuteAction = false` result cached before (or racing with) the on-chain grant was then served for the full **3600s `time_to_live`** (`CACHE_TTL_SECS` in `blockchain_cache.rs`), which matches the observed ~1-hour outage exactly. The sibling actions worked because their cached results were `true`.

No invalidation hook can ever cover out-of-band on-chain mutations, so the fix has to be at the cache policy level.

## Fix

Per-entry expiry via `moka::Expiry` on the permission caches:

- **authorized (`true`) results**: unchanged, 60-minute TTL
- **denied (`false`) results**: expire after **30 seconds**

The `(canExecute, canUseWallet)` pair cache treats any `false` in the pair as a denial. Wallet derivation lookups are unchanged (lookup errors were never cached by `try_get_with`). Also drops the redundant `time_to_idle` on the permission caches — with `tti == ttl`, the hard `time_to_live` cap always fired first, so behavior for positive results is identical.

Worst case for a fresh on-chain grant is now ~30s of stale 403 instead of an hour, while repeated unauthorized requests still only hit the contract once per 30s per (key, CID).

## Testing

- New unit tests for the expiry policy (create + update paths, pair-cache truth table) and an end-to-end moka test that a denied entry expires while a granted one persists.
- `cargo test` (250 tests), `cargo clippy --all-targets`, `cargo fmt` — all clean.

### DIFF
--- a/lit-api-server/src/accounts/blockchain_cache.rs
+++ b/lit-api-server/src/accounts/blockchain_cache.rs
@@ -5,9 +5,14 @@
 //! so that repeated calls for the same API key and relevant parameters
 //! avoid redundant contract calls.
 //!
-//! TTL: 60 minutes idle (extending on every access) with a hard upper bound
-//! of 60 minutes from insertion (`time_to_live`), ensuring entries are never
-//! served beyond the TTL even under continuous traffic.
+//! TTL: permission results use per-entry expiration — positive (authorized)
+//! results live 60 minutes from insertion, while negative (denied) results
+//! expire after 30 seconds. ChainSecured accounts mutate permissions by
+//! sending wallet-signed transactions directly to the chain, which this
+//! server never observes, so no invalidation hook can cover those writes; a
+//! short negative TTL bounds how long a stale denial is served (previously a
+//! newly group-permitted action returned 403 for up to an hour).
+//! Wallet derivation lookups keep the flat 60-minute TTL.
 //!
 //! Invalidation uses a **per-account generation counter**: each API key hash
 //! has an associated generation number embedded in the cache key. Bumping the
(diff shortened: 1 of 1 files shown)


clawdbot-glitch003:
## Codex adversarial review (gpt, high reasoning) — findings & dispositions

| # | Severity | Finding | Disposition |
|---|----------|---------|-------------|
| 1 | High | Stale **grants**: cached `canExecuteAction = true` survived on-chain revocation for the full 1h TTL — the more dangerous stale direction for an authorization cache | **Fixed** (09f50c91): positive TTL capped at 300s |
| 2 | High | Generation-counter invalidation is process-local; in multi-replica deploys, a mutation handled by one replica never invalidates the others | **Bounded, documented**: replicas now converge within 300s (grants) / 30s (denials). A proper fix needs shared invalidation infra — out of scope here |
| 3 | Medium | Unbilled RPC amplification: repeated denied executes force a fresh chain read every 30s per (key, CID); varying CIDs bypass the cache entirely | **Accepted**: the varying-CID vector is pre-existing and TTL-independent (a fresh CID was always a miss); same-CID repeats cost 1 coalesced read per 30s per key |
| 4 | Medium | RPC errors after negative expiry aren't dampened (`try_get_with` doesn't cache errors) | **Accepted**: pre-existing — errors were never cached before this PR either |
| 5 | Medium | Wallet derivation cache had the same unobserved-mutation exposure at 1h | **Fixed** (09f50c91): derivation TTL also capped at 300s |

Codex confirmed no moka `Expiry` API misuse (default `expire_after_read` preserves the remaining absolute expiry; reads don't extend entries).

Net TTL policy after this PR: **grants ≤ 300s stale, denials ≤ 30s stale**, in every replica, regardless of whether the mutation went through this server or directly on-chain.

PR #490, opened 2026-06-12
DX audit fixes: JSON API errors, correct status codes, never charge failed requests

clawdbot-glitch003:
## Why

A live DX audit of the production quickstart (fresh account, README steps verbatim) found that the first error a new developer hits is unparseable and misleading:

- Every error was Rocket's **default HTML page** — the README's own `curl | jq` quickstart crashes on it.
- An **invalid API key returned `402 Payment Required`** — a typo'd key told you to pay.
- Malformed bodies also 402'd (billing guard ran before validation), and funded accounts were **charged $0.01 for requests that then failed validation**.
- Production `/version` reported `v1.1.7-modified` — a spurious dirty-build marker, awkward for a verifiable-compute product.
- Docs gaps: no error reference anywhere, the API page's section numbering jumped 1→3, the README didn't mention the ~15s on-chain first call or the crypto funding options the docs site already has.

## What changed

**API behavior (lit-api-server):**

| Before | After |
|---|---|
| HTML error pages | JSON `{error, message, fix, docs_url}` on every error; 402 states amount needed, balance, and funding paths |
| Invalid key → 402 (or 400/500 on billing endpoints) | → **401** (typed `UnknownApiKey`, preserved through the cache) |
| Stripe/RPC outage → 402 | → **503** ("nothing was charged, retry") |
| $0.01 management charge taken in the guard, before validation | Guard only **checks** credit; a response fairing settles the charge **after success**. Failed requests are never charged |
| `GET /create_wallet` (metered write on a replayable verb) | `POST /create_wallet` added; GET kept, marked deprecated in the spec |
| No way to try the API without a card | `STARTER_CREDITS_CENTS` (default **0 = off**, no behavior change) grants new accounts a starter balance at creation. Turning it on is a per-node config decision |

**Build:** `.dockerignore` excluded a tracked file (`examples/lit-solver-vault/dashboard/.env.local.example`), so in-image `git describe --dirty` always saw a deletion → the `-modified` suffix. Re-included; `-modified` now means actually-modified.

**Docs (hosted site ships from `docs/`):** new [Errors reference](docs/management/errors.mdx) (status table, billing guarantees, common failure sequences) wired into nav and linked from quickstart/api_direct/README; api_direct gets its missing "2. Add funds" section, the 15-second note, and POST create_wallet; README/quickstart truth pass (credits framing, crypto/LITKEY options, Account Master Wallet).

**Repo onboarding:** CONTRIBUTING.md (complete prereq list, repo map, spec/k6 regen workflow), CHANGELOG.md, full `.env.example`, rewritten stale `lit-api-server/README.md` (it documented removed `/transfer/v1` + `/swaps/v1` APIs), new READMEs for lit-billing-core + lit-static, expanded lit-core, fixed `just k6-test` → `just test` in deployment docs.

## Notes for review

- `spec.json` + `k6/litApiServer.ts` regenerated per the `k6-client-check` CI gate (spec diff includes some pre-existing drift since the committed spec was last generated in May).
- **No deploy-pipeline changes**: compose file, workflows, and justfiles untouched; starter credits ship disabled.
- Settlement semantics: the guard→fairing window means concurrent requests can slightly overdraw — same staleness class already accepted for the balance cache (CPL-246). If settlement fails after a success response, the op stands and it's logged + counted (`billing.charge.settlement_failed`).
- k6 correctness suites only pin 401-for-missing-header, which is unchanged; the api-key-security suite should be re-run against a live stack to confirm the new 401-for-invalid-key path.

## Testing

- `cargo test`: 237 passed, 0 failed (includes 4 new catcher tests: JSON body, status preserved, guard detail override, JSON 404).
- `cargo clippy --all-targets`: clean.
- `cargo run --bin openapi_spec` verified `/create_wallet` has `get` (deprecated) + `post`.
- Not yet run: k6 suites against a live stack (needs deployed env); recommend a staging smoke before cutover.


clawdbot-glitch003:
Rebased onto current main (5 commits landed mid-flight, including auto-topup #448 and contract-revert decoding #481). Integration commit d9c3f40d aligns the two: decoded `AccountDoesNotExist` reverts classify as invalid-key (401) in both the guards and the billing endpoints, and the #481 tests that pinned the old 400 now expect 401. Auto-topup is unaffected — it's driven by lit-payments via the internal endpoint, not by `charge()`. Full suite: 235 passed / 0 failed, clippy clean, spec + k6 client regenerated with no further drift.

PR #496, opened 2026-06-15
docs(plans): shrink lit-api-server Stripe surface to debit-only

clawdbot-glitch003:
## Summary

Adds `plans/stripe-billing-improvements.md` proposing two related changes:

- **Shrink the TEE's Stripe surface to zero HTTP endpoints.** Move payment-intent creation, confirmation, customer/email management, `GET /billing/balance`, `GET /billing/stripe-config`, webhooks, and reporting out of `lit-api-server` and into `lit-payments`. The TEE keeps only internal balance-read + debit on the charge path.
- **Redesign the balance cache as advisory-only.** Insufficient-balance cache hits become cache misses (with a short negative cache as anti-DoS), so external top-ups (auto top-up, admin credit, LITKEY) self-heal within a single request instead of waiting on TTL. Removes the cross-replica correctness bug that blocks horizontal scaling.

Phase 1 (cache fix) is shippable on its own and unblocks auto top-up correctness today.

## Test plan

- [ ] Review for completeness — is anything else in `stripe.rs` worth keeping in the TEE that this plan moves out?
- [ ] Confirm the wallet→customer lookup recommendation (Option A: TEE queries Stripe directly) is the right tradeoff vs. coupling to lit-payments.
- [ ] Confirm Phase 1 cache rules don't introduce a Stripe rate-limit risk for spam loops (negative-cache window sized appropriately).

### DIFF
--- a/plans/stripe-billing-improvements.md
+++ b/plans/stripe-billing-improvements.md
@@ -1,0 +1,171 @@
+# Stripe Billing — Shrink the TEE Surface
+
+Status: design proposal
+Author: Chris (with Claude)
+Date: 2026-06-15
+
+## TL;DR
+
+`lit-api-server` (which runs inside a TEE) currently owns the entire Stripe lifecycle: customer creation, email registration, PaymentIntent creation, payment confirmation + credit, balance caching, and credit deduction. This grew organically because `lit-payments` didn't exist yet.
+
+Now that `lit-payments` is a real backend, the TEE should own **only the two things that actually need to be inside a TEE**, both as internal operations on the debit path:
+
+1. **Read** a customer's Stripe balance (to gate a request before doing the work).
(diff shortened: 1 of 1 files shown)


GTC6244 (approved):
Plan looks good. 

It doesn't mention anywhere that the Stripe publishable key should be the only one used in LitPayments - this is actually a good angle, as we should be able to keep the Stripe private key inside the TEE ( the only place where we debit from accountns ).

glitch003:
> Plan looks good.
> 
> It doesn't mention anywhere that the Stripe publishable key should be the only one used in LitPayments - this is actually a good angle, as we should be able to keep the Stripe private key inside the TEE ( the only place where we debit from accountns ).

but i think lit-payments needs the stripe secret key too, so that it can credit on auto top up and also for litkey payments.

PR #499, opened 2026-06-15
feat(accounts): per-account permission cache invalidation + on-chain listener

clawdbot-glitch003:
## Why

Alternative design to #491. Keeps the permission cache fresh when on-chain account/permission state changes, without #491's per-mutation usage-key enumeration (the `list_api_keys` fan-out and 1000-key cap), and without flushing more than the account that actually changed.

The hard part of #491 is mapping an on-chain event to the exact cache entries it touches. That's only hard because cache entries are keyed by the *calling* key (master or usage), while the permission-determining state lives at the *account* level — so an account mutation has to reach N usage keys. This PR dissolves that mapping with a single insight: **everything under one account resolves to the same wallet address**, so we key invalidation by the wallet, not by the calling key.

## The design

**Cache = verdict memoizer.** The key is the contract call: `(fn, api_key_hash, params) → verdict`. The server models nothing about groups/scopes — the contract resolves those when you call `canExecuteAction(hash, cid)`. A miss just asks the chain.

**Invalidation = one generation counter per account, keyed by wallet.** Every cache key embeds the account's generation. `getAccountWalletAddress(hash)` resolves any key — master or usage — to the same account wallet, so:
- Bumping `account_gen[wallet]` invalidates the master **and every usage key** under the account at once. **No enumeration, no `list_api_keys`, no 1000-key cap.**
- A mutation to account A flushes **only** A. An attacker spamming their own account's writes can't flush anyone else's cache.

The wallet for a calling key is resolved lazily and memoized in a resolution cache, so the hot path does at most one `getAccountWalletAddress` call per key per TTL.

**The listener decodes just enough.** It polls `eth_getLogs` for the 15 `WritesFacet` signatures (which also serve as the topic0 filter, excluding billing events), decodes each matching log to its account `apiKeyHash`, dedupes per poll, and invalidates that one account. One drift-guarded macro is the single source of truth for both the filter set and the decode dispatch.

**TTL is a backstop.** Invalidation keeps the cache fresh; the TTL only covers what invalidation can't (missed-block gaps, an ownership transfer that moves the wallet before its resolution entry expires, a replica between boot and first poll). Denials use a short 30s TTL vs the 300s grant TTL, so a permission granted while the listener is lagging/down still becomes visible quickly.

## Why this is better than the alternatives

- **vs #491:** no usage-key enumeration, no `list_api_keys` chain call per mutation, no 1000-key cap, no O(n) fan-out. Bumping one wallet generation covers all usage keys lazily.
- **vs a global flush:** blast radius is one account, not all tenants. A cheap stream of attacker-owned on-chain writes degrades only the attacker's own cache, not everyone's RPC load. (This was the key finding from an adversarial review of an earlier global-counter version of this PR.)

## Stateless / horizontally scalable

- Generation map is process-local; each replica runs its own listener and converges off the same chain logs. No shared state.
- Cold boot: empty cache → all miss → all fresh; listener anchors at the current block, never needs history.
- Better for **ChainSecured** accounts than any write-path hook: the listener sees on-chain mutations regardless of which replica or wallet made them. Write-path hooks additionally give instant invalidation (no poll lag) for mutations this process performs.

## Known follow-ups (out of scope here)

- **Listener robustness (#5/#6 from review):** reorg handling (poll has no confirmation lag) and per-poll block-range capping (a long outage can grow the range past provider limits). Shared with #491's listener; worth a hardening pass.
- **Ownership transfer:** changes the account wallet; cached entries under the old wallet serve until the resolution-cache entry expires (≤ TTL). Rare and typically permission-neutral.

## Test coverage

- `blockchain_cache.rs` — 17 tests: per-account generation increment/wrap/isolation, key formats embed the generation, hit/miss across all four caches, **invalidation scoped to one account** (no global flush), **master and usage keys share one account generation**, and the positive/negative TTL expiry policy (denials expire before grants).
- `account_events.rs` — 5 tests: signature set complete (15) and unique, account-hash extraction from both `apiKeyHash` and `accountApiKeyHash` events, billing event ignored, drift guard (every filtered signature decodes to an account hash).

## Test plan
- [x] `cargo build -p lit-api-server` — clean
- [x] `cargo test --lib` — 227 passed, 0 failed
- [x] `cargo clippy` — clean
- [x] `cargo fmt -- --check` — clean

### DIFF
--- a/lit-api-server/src/account_events.rs
+++ b/lit-api-server/src/account_events.rs
@@ -1,16 +1,16 @@
 //! On-chain account-mutation event listener.
 //!
 //! Polls the AccountConfig contract for the permission-mutating events emitted
-//! by `WritesFacet`. When it sees *any* of them in a block range, it flushes the
-//! permission cache by bumping the global generation
-//! ([`blockchain_cache::invalidate_all`]).
+//! by `WritesFacet`. For each event it extracts the account `apiKeyHash` the
+//! mutation touched and invalidates just that account's cached verdicts via
+//! [`crate::accounts::invalidate_account_by_hash`] (which resolves the hash to
+//! its wallet and bumps that account's generation — see
+//! [`crate::accounts::blockchain_cache`]).
 //!
-//! That's the whole job. Because invalidation is global (see
-//! [`crate::accounts::blockchain_cache`]), the listener does **not** decode logs
-//! or extract account/key hashes — it only needs to know *whether* a relevant
-//! mutation happened. The event-signature set is used purely as the
-//! `eth_getLogs` topic0 filter, so high-volume billing/config events on the same
-//! contract address don't trigger pointless flushes.
+//! Invalidation is per-account, so a single account's churn never flushes other
+//! accounts' caches. The listener decodes only enough of each event to recover
+//! its account hash; it does not enumerate usage keys (bumping the account's
+//! generation invalidates the master and all usage keys at once).
 //!
 //! This closes the staleness window for changes made outside this process —
 //! ChainSecured wallet-signed transactions sent directly to the contract, or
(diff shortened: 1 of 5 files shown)


glitch003:
closing - going with #491

PR #500, opened 2026-06-16
feat(lit-approvals): email-approval as an action library + Neon shared store (D6.1)

clawdbot-glitch003:
## What

Implements plan **D6.1**: moves the email-approval primitive **off** lit-api-server's in-memory `ApprovalService` and **into action logic** backed by an **untrusted shared store** (Neon over HTTP). This fixes the v1 bug where pending approvals were dropped on every horizontal scale-out / ping-pong deploy — the store now survives both, and the runtime never holds approval state.

Architecture (per the agreed design — lit-api-server fully out of approvals, page on the flows frontend, email sent from the action):

```
phase 1  approval ACTION (in-TEE)   requestApproval()  → id+OTP, write pending row to Neon, email link
   ↓
human    approval PAGE (flows)      recordSubmission() → record click + typed OTP  (cannot approve)
   ↓
phase 2  approval ACTION (in-TEE)   checkApproval()    → verify OTP vs TEE-keyed HMAC, SIGN attestation, consume
   ↓
consume  consuming ACTION (in-TEE)  verifyApproval()   → signature + approvalId + request_hash + expiry
```

One CID-bound key (`getLitActionPrivateKey()`) powers both the OTP HMAC (domain-separated) and the secp256k1 attestation signature — **no service holds a signing key** (this also resolves the v1 "service can forge" trust-boundary follow-up). The attestation format is unchanged (`email-approval-v1` / `secp256k1-sha256`), so it still verifies in the existing runtime op *or* in action JS.

## Why this is safe with an untrusted DB

Assume the adversary can **read + write every Neon row**. Encoded as tests (17 passing):

- **Cannot forge** — approval is proven by the signature; a fabricated row fails `verifyApproval` (they lack the key).
- **Cannot substitute** a real attestation onto another operation/approval — `approvalId` + `request_hash` are signed. *(This is why naïve encrypt-and-compare was rejected: encryption ≠ authentication.)*
- **Cannot brute-force the OTP** — only `HMAC(otpKey, id:otp)` is stored; `otpKey` is TEE-derived, never persisted.
- **Can only deny / DoS** — availability, not integrity.

Single-use is atomic on consume but **rollback-able by a DB admin**, so true anti-replay is pushed to the **execution-layer idempotency key** (bind `request_hash` to a chain nonce / venue `clientOrderId`). Documented in the README.

## Verification

`lit-approvals` in isolation: **17 unit tests pass** (the threat model *is* the test suite — forgery, substitution, OTP, binding, single-use, expiry, signer isolation), `npm run typecheck` clean, `npm run build` → ~90 KB IIFE + ESM.

## Status: stage 1 of N (draft)

This PR is **only the foundation library** + Neon schema. Deliberately separate, reviewable follow-ups:

1. **Remove** `ApprovalService` + the remote `sendEmail`/`requestEmailApproval`/`checkEmailApproval` ops from lit-api-server and the runtime (they become unused once actions own the flow).
2. **flows PR**: add the `/approvals/<id>` page that calls `recordSubmission` (writes click/OTP to Neon).
3. **Wire** the `cex-sweep-with-email-approval` example to the new action flow.
4. **Validate** the `neonHttpQuery` transport against live Neon on the dev deploy (the tested seam is the `NeonQuery` executor; the HTTP wire shape is best-effort until then).

Stacked on #483 (`feat/lit-venues-layer`) since it builds on the D6.1 plan note and the current ApprovalService.

### DIFF
--- a/lit-approvals/.gitignore
+++ b/lit-approvals/.gitignore
@@ -1,0 +1,3 @@
+node_modules/
+dist/
+*.log
--- a/lit-approvals/README.md
+++ b/lit-approvals/README.md
@@ -1,0 +1,101 @@
+# @lit-protocol/lit-approvals
+
+The email-approval primitive (plan D6/M3) re-implemented as **action logic** —
+no `ApprovalService` in lit-api-server. State lives in an **untrusted shared
+store** (Neon over HTTP), the attestation is **signed in-TEE with the approval
+action's CID-bound key**, and verification is **in-TEE action JS**. This fixes
+the v1 in-memory store, which dropped pending approvals on every horizontal
+scale-out / ping-pong deploy (plan D6.1).
+
+> Library code only — like `lit-venues`, it bundles to an IIFE for inlining
+> into a Lit Action. It is reachable from the runtime because Neon's serverless
+> SQL endpoint and Resend are plain HTTPS (`fetch` / `Lit.Actions.proxiedFetch`).
+
+## Architecture
+
+```
+phase 1  (approval ACTION, in-TEE)   requestApproval()  → issue id+OTP, write pending row to Neon, email link
+   ↓
+human    (approval PAGE, flows)       recordSubmission() → record click + typed OTP   (cannot approve)
+   ↓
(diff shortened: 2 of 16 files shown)


glitch003:
closing - using a new design

PR #502, opened 2026-06-16
feat(lit-actions): in-TEE proxied fetch op (Lit.Actions.proxiedFetch)

clawdbot-glitch003:
## What

Extracted from #483 — **just the egress primitive**. The rest of #483 (venue connectors, email-approval, examples) is being reworked into flows separately, so once this lands #483 can be closed.

Adds **`op_lit_proxied_fetch`**, exposed as **`Lit.Actions.proxiedFetch`** — outbound HTTP for a Lit Action that can egress through a per-request authenticated proxy. This lets an action reach a geo-blocked venue (e.g. Binance returns HTTP 451 from a US enclave IP) via a chosen non-US IP.

Egress happens in-process via `reqwest` — the same place Deno's `fetch` egresses from — over an HTTP CONNECT tunnel, so **TLS to the destination is end-to-end and the proxy never sees venue credentials or payloads**. Responses are streamed with a 10 MiB cap, and the call counts against the per-action fetch quota via `op_increment_fetch_count` in the JS wrapper. It's a purely local op — **no gRPC round-trip to lit-node, so no proto changes**.

## Changes

Touches `lit-actions` only:

- `ext/bindings.rs` — the op, a per-proxy `reqwest::Client` pool (capped at 16, keeps TCP+TLS warm), `split_proxy_credentials`, registration in the `lit_actions` extension.
- `ext/js/02_litActionsSDK.js` — `Lit.Actions.proxiedFetch` wrapper (fetch-quota enforcement + fetch-like response shape).
- `ext/Cargo.toml` / `Cargo.lock` — adds `reqwest` (pinned to match `lit-actions/server` exactly, so feature unification is a no-op and we reuse the process aws_lc_rs provider) and `serde`.
- `tests/it.rs` — `proxied_fetch` integration test.

## Reviewer notes

- `#[op2(async(lazy))]`, not bare `async` — edition 2024 rejects the bare `async` keyword in attribute position.
- Proxy auth is applied via `.basic_auth()` because `reqwest::Proxy` doesn't forward URL userinfo as the `Proxy-Authorization` header.
- Request errors are **categorized only** (timeout/connect/request/body flags); the reqwest error is never interpolated because its `Display` embeds the request URL, and for signed venue calls the HMAC signature + order params ride in the query string.
- `tests/it.rs::proxied_fetch` is `#[ignore]`d (needs real outbound network, like `import_rewrite_cdn`): `LIT_VENUES_TEST_PROXY=… cargo test -p lit-actions-tests --test integration proxied_fetch -- --ignored`.

## Verification

- `cargo check -p lit-actions-ext` and `-p lit-actions-tests --tests` both green.
- `cargo test -p lit-actions-tests --test integration -- fetch` → `fetch` passes, `proxied_fetch` correctly listed as ignored.


lit-actions/ext/bindings.rs
```diff
@@ -263,6 +263,224 @@ pub async fn op_update_resource_usage_external(
...
+#[derive(serde::Deserialize)]
+struct ProxiedFetchRequest {
+    url: String,
+    #[serde(default = "default_proxied_fetch_method")]
+    method: String,
+    #[serde(default)]
+    headers: Vec<(String, String)>,
+    #[serde(default)]
+    body: Option<String>,
+    /// `http(s)://[user:pass@]host:port`. Empty/None → a direct (unproxied) request.
```
GTC6244: Any reason to allow an Empty/None here ?   As far as I can tell it probably would only throw statistics off a bit ....  do we care ?

lit-actions/ext/bindings.rs
```diff
@@ -263,6 +263,224 @@ pub async fn op_update_resource_usage_external(
...
+    #[serde(default)]
+    proxy: Option<String>,
```
GTC6244, line 282: Actually - Claude tells me this a secured whole that would allow direct egress to the enclave - still requires some insider knowledge, but I think we need to block this.

>>>

Beyond the stats/quota accounting, the meaningful problem is silent fallback to direct egress.

In op_lit_proxied_fetch (bindings.rs:411-414) the proxy is .map(str::trim).filter(|p| !p.is_empty()), so an empty or whitespace-only proxy string collapses to None → a direct request from the enclave’s own IP.

The entire reason this op exists is that the enclave IP is geo-blocked (451 from Binance) and credentials/origin must egress through a chosen non-US IP. If the proxy value is ever missing or malformed (config bug, empty env var, bad string interpolation in an action), the call doesn’t fail — it silently sends the signed request, API keys, and HMAC out the front door from the enclave IP. That’s the exact thing the primitive is supposed to prevent. For a security-sensitive egress primitive this should fail closed: if proxy was supplied but resolves to empty, error rather than degrade to direct.

If a true unproxied mode is wanted, make it an explicit, separate affordance (e.g. proxy: null deliberately = direct, but a non-null-but-empty string = error) so misconfiguration can’t masquerade as “direct is fine.”

lit-actions/ext/bindings.rs
```diff
@@ -263,6 +263,224 @@ pub async fn op_update_resource_usage_external(
...
+/// per-request proxy strings).
+const PROXIED_FETCH_CLIENT_POOL_MAX: usize = 16;
+
+lazy_static::lazy_static! {
+    /// One `reqwest::Client` per proxy URL ("" = direct). `Client` is an Arc
+    /// around a connection pool, so reuse keeps TCP+TLS sessions warm across
+    /// calls instead of paying a fresh handshake per request (plan M2
+    /// follow-up). Keys contain proxy credentials — they are secret material
+    /// and must never be logged.
+    static ref PROXIED_FETCH_CLIENTS: std::sync::Mutex<std::collections::HashMap<String, reqwest::Client>> =
```
GTC6244: RwLock probably fits better here - faster under concurrency, and more idiomatic across our codebase.   Not a huge blocker, though.

lit-actions/ext/bindings.rs
```diff
@@ -263,6 +263,224 @@ pub async fn op_update_resource_usage_external(
...
+                "op_lit_proxied_fetch: response exceeded {PROXIED_FETCH_MAX_BYTES} bytes"
+            )));
+        }
+        buf.extend_from_slice(&chunk);
+    }
+
+    Ok(ProxiedFetchResponse {
+        status,
+        headers,
+        body: String::from_utf8_lossy(&buf).into_owned(),
```
GTC6244, line 505: What if we're requesting binary data?   It'll be corrupted on the from_utf8_lossy ....

GTC6244 (changes requested):

lit-actions/ext/bindings.rs
clawdbot-glitch003: Good catch — this turned out to be the substantive one (see the thread below). Fixed in 85972484: a supplied-but-empty/whitespace proxy now fails closed (errors) instead of trimming to `None` and silently going direct. The quota/stats angle you noted is real but minor; the egress fallback was the actual problem.

lit-actions/ext/bindings.rs
clawdbot-glitch003: Agreed, and fixed in 85972484. The decision is now in a small `resolve_proxy()` helper with exactly the semantics you described:

- omit field / `null` = deliberate direct request (allowed)
- `Some(non-empty)` = proxied (trimmed)
- present-but-empty/whitespace = **error** (fail closed)

So a config bug (empty env var, bad interpolation) can no longer masquerade as "direct is fine" and leak the signed request + keys + HMAC out the enclave IP. Covered by unit tests (`resolve_proxy_none_is_direct`, `_present_is_trimmed`, `_supplied_but_empty_fails_closed`).

lit-actions/ext/bindings.rs
clawdbot-glitch003: Done in 85972484 — switched to `RwLock` for codebase consistency. Perf is a wash in practice (the pool saturates at <=16 entries almost immediately and is read-only after that, and the client build happens outside the lock), but the idiomatic-consistency point is fair so I took it.

lit-actions/ext/bindings.rs
clawdbot-glitch003: Correct — `from_utf8_lossy` mangles non-UTF-8 (binary/gzip/protobuf) responses. For v1 I documented this as a text-only contract (Rust response struct + the JS `proxiedFetch` JSDoc) rather than changing behavior: the body crosses into JS as a string and every current consumer is a JSON/text venue REST API, so it round-trips exactly today. Proper binary support (base64 envelope or a `responseType` flag) is a follow-up — flagged in the code comment at the `from_utf8_lossy` site. Shout if you want the base64 path in this PR instead.

GTC6244 (approved):

PR #521, opened 2026-06-23
perf(lit-actions): warm secp256k1 precompute into the V8 snapshot

clawdbot-glitch003:
## What

Signing Lit Actions (`soak_ecdsa_sign`, ~248ms avg) run ~2.5× slower than encrypt/decrypt (~101ms avg) in the soak test, despite signing here being a *local* `ethers.Wallet.signMessage()` — not threshold MPC.

Root cause: **ethers 5.7 (elliptic.js) builds its secp256k1 base-point precompute table lazily, on the first EC op in an isolate.** Evaluating the library into the startup snapshot (which `99_patches.js` already forces via its import) doesn't trigger this — only an actual key-derivation/sign op does. Since every Lit Action runs in a **fresh, one-shot isolate** (`worker_pool.rs`: isolates are pre-warmed from the snapshot but each serves exactly one request, then is dropped), every single signing request recomputes that table from scratch. Encrypt/decrypt never touch the curve, so they never pay it — that's the entire gap.

## Fix

One synchronous throwaway `SigningKey + signDigest` at **snapshot-build time** (end of `99_patches.js`). The precompute table hangs off the curve singleton (reachable from the live ethers module), so it's serialized into the V8 startup snapshot and **every isolate boots warm** — zero added cost on the request path or at worker bootstrap.

## Verification

Measured in a fresh isolate through the real snapshot runtime (temporary harness, not committed) — ethers-sign vs a no-ethers control, one fresh isolate per request:

| | control (no ethers) | ethers sign | **sign − control** |
|---|---|---|---|
| **baseline** (no warmup) | 17.6 ms | 57.1 ms | **39.5 ms** |
| **with warmup** | 18.7 ms | 31.4 ms | **12.7 ms** |

The precompute **survives snapshot serialization** and is removed from the per-request path. Numbers are small here (debug build, M-series CPU, no network); the residual ~12.7ms is just the warm ethers ops. On the production TEE CPU (slower + hardened V8 with memory-protection-keys/clear-free-memory) the cold precompute is the ~120ms floor gap seen in the soak (`sign` min 206ms − `encrypt` min 87ms ≈ 119ms).

Expected impact: `soak_ecdsa_sign` p95 should drop from ~274ms toward the encrypt/decrypt range, since the warmup removes the one thing that distinguishes them.

Full `lit-actions` integration suite passes (27 passed / 0 failed / 3 ignored). The warmup is intentionally placed at the **end** of `99_patches.js` so it doesn't shift `__litEvalCached`'s line number, which two stack-trace tests pin.

## Notes / follow-up

- `viem` is **not** primed here — its bundle uses lazy esbuild module initializers and isn't exposed as a runtime global, so the same trick doesn't apply cleanly. Worth a follow-up for viem-based signing actions.
- Best-effort `try/catch`: an ethers API drift can never break the snapshot build; worst case we fall back to today's per-request precompute.

### DIFF
--- a/lit-actions/ext/js/99_patches.js
+++ b/lit-actions/ext/js/99_patches.js
@@ -64,41 +64,52 @@ globalThis.__litEvalCached = (source, specifier) => {
 // existing code that expects it to be available
 globalThis.global = globalThis;
 
-// Prime ethers' secp256k1 base-point precompute table at snapshot-build time.
+// Prime ethers' secp256k1 base-point precompute at snapshot-build time.
 //
-// ethers 5.7 (elliptic.js) builds a windowed base-point multiplication table
-// lazily, on the FIRST EC operation in an isolate. Merely evaluating the
-// library (the import at the top of this file) does not trigger it -- only an
-// actual key-derivation/signature op does. Because every Lit Action runs in a
-// fresh, one-shot isolate, an un-primed snapshot makes every signing request
-// pay that precompute again (~100-150ms on the production TEE CPU; measured
-// ~45-50ms cold vs ~2ms warm locally on the runtime's exact bundle). Doing one
-// throwaway op here runs it ONCE at snapshot-build time; the resulting table
-// hangs off the curve singleton (reachable from the live ethers module) and is
-// serialized into the startup snapshot, so every isolate boots warm. Verified:
-// in a fresh isolate the ethers-sign extra cost drops from ~40ms to ~13ms.
+// ethers 5.7 (elliptic.js) lazily builds a windowed base-point multiplication
+// table the first time the secp256k1 curve is constructed (elliptic's EC
+// constructor calls g.precompute()). Evaluating the library into the snapshot
+// (the import at the top of this file) does NOT construct that curve -- only an
+// actual key/signature op does, and ethers caches the curve in a module-level
+// singleton on first use. Because every Lit Action runs in a fresh, one-shot
+// isolate, an un-primed snapshot makes the first signing op in EACH isolate
+// rebuild the table (~100-150ms on the production TEE CPU; measured ~45-50ms
+// cold vs ~2ms warm locally on the runtime's exact bundle). Constructing one
+// SigningKey here forces that curve + precompute ONCE, at snapshot-build time;
+// the cached curve is serialized into the startup snapshot so every isolate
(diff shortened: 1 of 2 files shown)


clawdbot-glitch003:
## Adversarial review (Codex) + resolutions

Ran an independent adversarial pass (OpenAI Codex, read-only, high reasoning) against the diff. It read the ethers bundle internals and the Rust snapshot wiring. Summary of findings and how they're addressed in the follow-up commit:

- **P1 — blanket `try/catch` could silently no-op** (ethers API drift renaming `SigningKey`/`signDigest`, or the op throwing → build stays green, every isolate reships the cold precompute, zero signal). **Fixed:** the warmup now records a build-time marker `globalThis.__litSecp256k1Warmed`, true only if the sign path ran end-to-end and produced a signature. New deterministic test `secp256k1_precompute_warmed_in_snapshot` asserts it's true in a fresh isolate → silent failure becomes a loud CI failure. Catch kept graceful so an unrelated ethers bump degrades instead of breaking the snapshot build.

- **P2 — precompute table now in every worker snapshot, no size/memory number.** **Measured** (build A/B, warmup on vs off): `14,332,137` vs `14,111,081` bytes = **+221,056 bytes (~216 KB, 1.56% of the 14 MB snapshot; ~0.3% of the 64 MB per-isolate heap)**. That's the secp256k1 windowed-NAF table. Bounded, constant, paid once per isolate at deserialize. Acceptable.

- **P2 — comment overstated `signDigest`'s role.** **Fixed:** comment now attributes the persistent win to the base-point precompute from the `SigningKey`/`EC` construction; `signDigest` is belt-and-suspenders (also exercises the deterministic RFC-6979 sign path, and lets us set the success marker).

- **P3 — `__litEvalCached` line pin was a maintainability trap.** **Fixed:** de-pinned the volatile `99_patches.js:LINE:COL` in the `reference_error`/`throw_error` stack-trace assertions (`normalize_patches_line`). The frame and full user-code stack are still asserted exactly.

**Codex explicitly cleared** the two highest-risk design questions: no snapshot nondeterminism (ECDSA here is RFC-6979 deterministic, no entropy enters the snapshot) and no cross-request state leak (one-shot isolates).

Integration suite after the changes: **28 passed / 0 failed / 3 ignored.**

Open follow-up (unchanged): viem isn't warmed — its bundle uses lazy esbuild initializers and isn't a runtime global, so it needs a different hook.

PR #556, opened 2026-07-03
fix(billing): stop guard from cache-poisoning on Stripe Search index lag

### LINKED ISSUE #555: Billing guard races Stripe Search index: duplicate zero-credit customer cached, 402 on first action
## Summary

The `/lit_action` billing guard looks up the Stripe customer by `metadata.wallet_address` using **Stripe Customer Search**, which is eventually consistent (new customers take ~15–30s to become searchable). If a billable action arrives before the index catches up, the guard misses the funded customer, **creates a duplicate zero-credit customer, caches it, and returns 402** — a poisoned state, not a transient failure.

## Impact

Any consumer that funds a customer and calls `lit_action` shortly after hits this. In Flows we had to add a defensive polling loop (up to 20 × 1.5s polls of `stripe.customers.search`) after writing the signup credit, waiting for the customer to become search-indexed before letting the user proceed. That wait is the dominant cost of our `mint-usage-key` endpoint (~25s observed on prod) — real signup/venue-connect latency caused entirely by this guard design.

## Root cause

Stripe Search is explicitly documented as not read-after-write consistent ("data is searchable within a minute"). Using it as the authoritative customer lookup on the billing hot path means a freshly created + funded customer is invisible to the guard for tens of seconds, and the miss path makes the failure permanent by creating and caching a duplicate customer with zero balance.

## Suggested fixes (any one kills the race)

1. **Create the Stripe customer at account/key creation time** (e.g. in `add_usage_api_key_with_signature` / ChainSecured account creation), store the customer ID, and have the guard look up by ID instead of Search. Cleanest: the customer exists and is indexed long before the first billable call.
2. **Keep a durable wallet→customer mapping** in Chipotle's own store instead of querying Search on the hot path.
3. At minimum, **don't cache-poison on a miss**: treat "customer not found in Search" as retryable (or re-check with a direct read) rather than creating a duplicate zero-credit customer.

## Workaround we're running today (Flows)

After granting the signup credit we poll `customers.search` until the customer appears (up to ~30s) before completing signup, so the guard can't race the index. This blocks the user-facing `mint-usage-key` response for ~25s on prod. We can move the wait around, but the race is only truly fixable on the Chipotle side.

clawdbot-glitch003:
Fixes #555.

## Problem

Stripe Customer Search is eventually consistent — a freshly created + funded customer can be invisible to `customers/search` for tens of seconds. The `/lit_action` billing guard did find-or-create on that search, so a billable call arriving before the index caught up **created a duplicate zero-credit customer, cached it, and returned 402** — a poisoned state, not a transient failure. Flows works around this today with a ~25s polling loop in `mint-usage-key`.

## Fix

Two independent changes; either alone kills the duplicate-create race (this is suggested fix 3 from the issue, plus a hardening layer):

**1. Billing guards never create customers** (`lit-api-server/src/stripe.rs`)
- New `find_customer_by_wallet`: cache → search, find-only. A miss returns `Ok(None)` — nothing created, nothing cached (no negative caching), so the next request re-checks Stripe and heals as soon as the index catches up (≤ ~1 min).
- `check_credit` and `charge` use it. "No customer" → retryable `402 InsufficientCredits { available: $0 }`, which is also the semantically correct answer for a genuinely unfunded wallet.
- Legitimate creation paths keep find-or-create: account creation (which eagerly creates the customer and warms the same in-process cache the guard reads), top-ups, confirm-payment, balance display, email registration.

**2. Customer creation is idempotent per wallet** (`lit-billing-core/src/customer.rs`)
- `find_or_create_by_wallet` now POSTs with a deterministic `Idempotency-Key` (`customer-create-{wallet}`). Two creators that both miss the index — api-server replicas, retried account creation, concurrent top-up + signup — get the *same* customer back from Stripe's 24h idempotency replay instead of minting duplicates. The 24h window comfortably covers the ~1-minute search-index lag.

## Behavior changes

- A funded-but-not-yet-indexed customer now gets a transient, self-healing 402 instead of a permanent one. Flows can drop (or drastically shorten) its 20×1.5s search-poll: the worst case for an action fired within seconds of funding is a retryable 402, not a poisoned account.
- A wallet that never funded no longer gets a zero-credit Stripe customer minted by the guard; the customer is created at account creation or first top-up instead. Nothing depended on guard-side creation (starter credits are only granted in `new_account`, which creates its own customer).

## Testing

- `cargo test` in `lit-billing-core` (73 passed, includes new idempotency-key contract test)
- `cargo test --lib` + `cargo clippy --all-targets` clean in `lit-api-server`; `cargo check` clean in `lit-payments`

### DIFF
--- a/lit-api-server/src/core/v1/guards/billing.rs
+++ b/lit-api-server/src/core/v1/guards/billing.rs
@@ -93,11 +93,24 @@ fn billing_error_status(request: &Request<'_>, e: BillingError) -> Status {
                     cents_to_display(required_cents),
                     cents_to_display(available_cents),
                 ),
-                format!(
-                    "Add funds (minimum $5.00, card or crypto) in the dashboard at \
-                     {DASHBOARD_URL} or via POST /core/v1/billing/create_payment_intent. \
-                     Check your balance with GET /core/v1/billing/balance."
-                ),
+                if available_cents == 0 {
+                    // A $0 balance can also mean the customer was funded moments
+                    // ago and Stripe's search index hasn't caught up yet (#555)
+                    // — that case is transient and heals within about a minute.
+                    format!(
+                        "Add funds (minimum $5.00, card or crypto) in the dashboard at \
+                         {DASHBOARD_URL} or via POST /core/v1/billing/create_payment_intent. \
+                         If you added funds within the last minute, retry shortly — new \
+                         billing accounts can take up to a minute to become visible. \
+                         Check your balance with GET /core/v1/billing/balance."
+                    )
+                } else {
+                    format!(
+                        "Add funds (minimum $5.00, card or crypto) in the dashboard at \
+                         {DASHBOARD_URL} or via POST /core/v1/billing/create_payment_intent. \
+                         Check your balance with GET /core/v1/billing/balance."
+                    )
+                },
             );
             Status::PaymentRequired
         }
--- a/lit-api-server/src/stripe.rs
+++ b/lit-api-server/src/stripe.rs
@@ -477,6 +477,12 @@ pub async fn resolve_wallet_address(api_key: &str, state: &StripeState) -> Resul
 
 /// Find the Stripe customer for this wallet address, creating one if none exists.
 ///
+/// Only for paths where creating a customer is legitimate (account creation,
+/// top-ups, email registration, balance display). The billing guards
+/// (`check_credit` / `charge`) must use [`find_customer_by_wallet`] instead:
+/// the underlying lookup is Stripe Search, which is eventually consistent, so
+/// creating on a search miss can duplicate a freshly funded customer (#555).
+///
 /// Results are cached in memory to avoid duplicate customer creation caused by
 /// Stripe Search API indexing lag (newly created customers may not appear in
 /// search results for several seconds).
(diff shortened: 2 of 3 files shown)


clawdbot-glitch003:
## Codex adversarial review (round 1)

Ran an adversarial second-opinion review (OpenAI Codex, high reasoning) against this diff. Six findings; three fixed in 2eec8748, one mitigated, two accepted as pre-existing/inherent:

| # | Finding | Verdict | Action |
|---|---------|---------|--------|
| 1 | Wallets that **already** have duplicate customers in prod (minted by the #555 bug) could still cache the zero-credit dup: search used `limit=1` with unspecified ordering | Real | Fixed — `find_by_wallet`/`find_summary_by_wallet` now fetch up to 100 matches and deterministically pick the **oldest** customer (the funded original) |
| 2 | Freshly funded users can still see a transient 402 on a cold replica until Search indexes (~15–30s) | Real but inherent | Accepted — this is exactly the issue's suggested fix 3 semantics: retryable, self-healing, never poisoned. Fully removing the window needs eager creation at key-mint or a durable wallet→customer store (follow-up) |
| 3 | Retryable index-lag 402 is indistinguishable from a genuine no-funds 402 | Partially real | Mitigated — the $0-balance 402 detail now says freshly funded accounts can take up to a minute to appear; a new error variant would break the client status contract for marginal gain |
| 4 | Stripe replays the first response under an idempotency key for ≥24h **including 500s** → one transient create failure pins the wallet for a day | Real (confirmed against Stripe docs) | Fixed — on create failure: re-check search, then retry once under a deterministic `-r2` fallback key (racers on the fallback still converge) |
| 5 | Find-only guard path = one Stripe Search per request for unfunded wallets → search rate-limit amplification | Real | Fixed — 5s-TTL per-wallet miss cooldown (bounded negative cache; delays recovery ≤5s, cannot poison) |
| 6 | Concurrent `charge()` calls can overdraw via the non-atomic cached-balance check/insert | Pre-existing | Out of scope — documented as acceptable overcharging bound per CPL-246; untouched by this diff |

Tests after fixes: `lit-billing-core` 73 passed, `lit-api-server` 253 passed + clippy clean, `lit-payments` checks clean.

clawdbot-glitch003:
## Codex adversarial review (round 2)

Ran a second adversarial pass (OpenAI Codex, high reasoning) focused on the round-1 hardening. Six findings; three real regressions fixed in the latest commit, one edge mitigated, two out-of-scope.

| # | Finding | Verdict | Action |
|---|---------|---------|--------|
| 1 | The `-r2` idempotency fallback **reintroduces the #555 duplicate race**: on `idempotency_key_in_use` (primary create still in flight elsewhere) or a post-create client timeout, falling back to a different key mints a second customer | Real regression | **Fixed** — dropped the fallback; single fixed key so racers converge and callers retry under the same key (replaying the stored success) |
| 2 | If both keys store a 5xx, creation is pinned for 24h anyway — fallback trades a rare failure for the common one | Real regression | **Fixed** — same as #1. Residual (stored 5xx pins one wallet's creation ≤24h) is accepted, far cheaper than duplicating on every signup race |
| 3 | `min_by_key(created)` has no tie-breaker → same-second duplicates fall back to Stripe's arbitrary search order | Real | **Fixed** — rank by `(balance, created, id)`; deterministic tie-break |
| 4 | "oldest == funded original" is not guaranteed (an older zero-credit dup + newer funded dup → permanent 402) | Real | **Fixed** — rank prefers the **funded** customer (credit = negative Stripe balance, already in the search hit → no extra API calls) |
| 5 | `limit=100` ignores search pagination; funded customer could be on an unfetched page | Real, extreme edge | **Mitigated** — `has_more` warning log (a >100-customer wallet is pathological; balance-ranking still picks the funded one from page 1) |
| 6 | `lit-payments` `auto_topup_config` is UNIQUE by wallet but upserts on `customer_id` — selection flip could hide/duplicate a row | Pre-existing, separate subsystem | **Flagged, not fixed** — latent on main. This PR makes selection *deterministic* (main's first-hit could flap request-to-request), so it improves that path's stability. Worth a dedicated fix |

Added a unit test covering funded-over-older, same-second id tie-break, and missing-field defaults.

Tests after fixes (rebased onto latest main incl. #554): `lit-billing-core` 74 passed, `lit-api-server` 253 passed + clippy clean, `lit-payments` 120 passed. `cargo fmt --check` clean on all three.

GTC6244 (approved):
Clean clever solution !

PR #560, opened 2026-07-07
docs(actions): plan multi-language Lit Actions feature gating + capabilities (CPL-345)

GTC6244:
Plans CPL-345: the feature-gating and discovery layer on top of the already-built gVisor any-language runner (PRs #557/#558), rather than a new runner. It defines the two methods — **Bundle** (already shipped as `/lit_binary_action`) and **Direct Coding** (a thin front-end that synthesizes a one-file bundle down the same path) — and a `GET /core/v1/configuration/get_supported_languages` endpoint returning language name, runtime version(s), and direct-coding/bundle support. The core design reconciles two layers of truth (measured rootfs image contents × a per-environment `languages.<env>.toml` policy) via a new `GetCapabilities` gRPC so advertised versions come from the real image and cannot drift. Deploy-YAML gating (parallel to `ARG NODE_CONFIG`, plus image selection governed by `compose_hash`) lets `next` expose Python while prod stays JS-only until sign-off. Ships as a plan doc only (`plans/multi-language-lit-actions.md`) with a phased rollout and open questions; no code changes.

### DIFF
--- a/plans/multi-language-lit-actions.md
+++ b/plans/multi-language-lit-actions.md
@@ -1,0 +1,213 @@
+# Multi-Language Lit Actions — Feature Gating & Capabilities (CPL-345)
+
+**Status:** Draft / plan
+**Linear:** [CPL-345](https://linear.app/litprotocol/issue/CPL-345/plan-for-supporting-multiple-languages-for-lit-actions-in-cvm)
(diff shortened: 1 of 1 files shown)


GTC6244:
closing instead of fixing as premise isn't quite right

PR #572, opened 2026-07-13
feat(lit-actions): Claude-in-startup-script example on the base rootfs (CPL-356)

GTC6244:
Adds a `claude` example bundle to both the gVisor runner and the local CLI, demonstrating PR #571's startup-script contract (the sandbox only runs `bash startup.sh`, so the script provisions its own toolchain). The `startup.sh` installs Claude Code onto the "Slimworm" base rootfs (debian:bookworm-slim, which has `curl` but no Node) via the native installer, then — when an `ANTHROPIC_API_KEY` top-level js-param is present — runs a headless `claude -p` prompt and records the reply with `lit set-response` (metering the call via `lit increment-fetch-count`); with no key it does an install-only run returning the version. The two `startup.sh` copies are near-identical (the local one warns it installs onto your machine and uses `uname -sm`), and both crate READMEs now list the example plus its network + API-key requirements. No test harness enumerates `examples/`, so this is never auto-executed in CI (it hits the network + installer).

**Base branch note:** this is stacked on the CPL-355 startup-script contract, so it targets `feature/cpl-355-startup-script-for-gvisor` (PR #571) rather than `main` — retarget to `main` once #571 lands.

### DIFF
--- a/lit-actions/gvisor-server/README.md
+++ b/lit-actions/gvisor-server/README.md
@@ -188,8 +188,9 @@ cargo test -p lit-actions-gvisor-server
 ```
 
 `image/` has the base-image Dockerfile; `examples/` has bundle examples —
-`shell`/`python` are minimal, and `claude` installs Claude Code from its
-`startup.sh` onto the base rootfs (debian:bookworm-slim) and runs a headless
+`shell`/`python` are minimal, and `claude` executes the official Claude
+Code installer script (`https://claude.ai/install.sh`) from its `startup.sh`
+onto the base rootfs (debian:bookworm-slim) and runs a headless
 prompt, demonstrating a startup script that provisions its own toolchain
 (needs sandbox network egress + an `ANTHROPIC_API_KEY` js-param).
 
--- a/lit-actions/gvisor-server/README.md
+++ b/lit-actions/gvisor-server/README.md
@@ -187,7 +187,11 @@ anywhere (macOS included):
 cargo test -p lit-actions-gvisor-server
 ```
 
-`image/` has the base-image Dockerfile; `examples/` has bundle examples.
+`image/` has the base-image Dockerfile; `examples/` has bundle examples —
+`shell`/`python` are minimal, and `claude` installs Claude Code from its
+`startup.sh` onto the base rootfs (debian:bookworm-slim) and runs a headless
+prompt, demonstrating a startup script that provisions its own toolchain
+(needs sandbox network egress + an `ANTHROPIC_API_KEY` js-param).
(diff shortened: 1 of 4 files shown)


GTC6244:
## Review: startup-script content ✅

Reviewed the two `startup.sh` bundles against the CPL-355 sandbox contract; no code changes needed. What I verified:

- **`$HOME` install target is valid.** The native installer drops `claude` in `$HOME/.local/bin`, and the runsc sandbox sets `HOME=/tmp` (`sandbox/runsc.rs`), a writable in-memory overlay — so `/tmp/.local/bin` works and `export PATH="$HOME/.local/bin:$PATH"` resolves correctly.
- **js-param → env injection matches the contract.** `ANTHROPIC_API_KEY` / `PROMPT` are valid, non-reserved env names (`RESERVED_ENV` = PATH/HOME/TMPDIR/LIT_OP_SOCK/LIT_ACTION_IPFS_ID), so they inject cleanly.
- **Exit semantics are correct.** `set -euo pipefail` means a failed install or a failed `claude -p` aborts non-zero (→ failed execution, no response), and the install-only path `exit 0`s with a recorded version. Response JSON is built with `jq` so an arbitrary `claude` reply can't break it.
- **No CI auto-exec.** No test harness enumerates `examples/`, so this bundle (which needs network + the installer) never runs in CI.

### Known runtime prerequisites (documented in both READMEs)
- Sandbox **network egress** to `claude.ai` (install) and `api.anthropic.com` (prompt).
- Base rootfs toolchain: `curl` + `jq` (both present in `Dockerfile.rootfs`); no Node needed by design.

GTC6244:
## CI: red checks were a transient GitHub Actions outage, not this diff

The initial run showed 5 failing checks (contract-bindings, forge-test, k6-client-check, lit-payments, migration-versions). Root cause is infrastructure, not the code — every failing job died at the action-resolution step:

```
##[error]Internal Server Error occurred while resolving "actions/checkout@v4"...
Failed to resolve action download info. Error: Service Unavailable
```

All 5 failed within the same ~13:14 UTC window while GitHub couldn't resolve `actions/checkout`, `setup-node`, `rust-toolchain`, `foundry-toolchain`, etc. Corroborating evidence:

- This PR touches only two `startup.sh` example scripts + two READMEs — nothing that could affect Rust, Foundry, migrations, or k6 codegen.
- The base branch (**PR #571**) is fully green on the identical stack.

**Resolution:** re-ran the failed jobs; `migration-versions` already flipped back to pass and the rest are re-running. No code change required.

Issue #575, opened 2026-07-14
Security: public derivation paths + per-account-only uniqueness let any account hijack another account's PKP wallet


clawdbot-glitch003:
## Summary

PKP derivation paths are **public on-chain**, but possession of a `(pkpId, derivationPath)` pair is effectively all that's needed to make the node re-derive and sign with that key. Because `registerWalletDerivation` only enforces uniqueness **within a single account** — not globally — a malicious ChainSecured account can register a *victim's* already-public wallet derivation into its **own** account and then drive the node's signing/encryption endpoints to operate with the victim's private key.

This is a pre-existing issue independent of the flows#532 "derived-wallet-address endpoint" work. The proposed new endpoint does **not** make it worse (paths and addresses are already fully public today), but the investigation for that work surfaced it, so filing separately.

## Why paths are public

`derivationPath` is exposed two ways on Base (prod `chain = base`, `NodeConfig.prod.toml`):

1. **Event data** — `registerWalletDerivation` emits `WalletDerivationRegistered(masterHash, pkpId, derivationPath)` (`WritesFacet.sol:691`). `derivationPath` is non-indexed but present in the log data, so it is trivially readable by anyone scanning logs.
2. **Public getter** — `getWalletDerivation(apiKeyHash, walletAddress)` is `public view` (`ViewsFacet.sol:158`) and returns `pkpData[walletAddress].id` (the path) for *any* account, given its `apiKeyHash` and a wallet address. `listPkps` (`ViewsFacet.sol:251`) exposes the same data paginated.

So both the wallet address and its derivation path are public knowledge for every registered PKP.

## Why that's dangerous: the key is a stateless function of the path

Key derivation is deterministic and stateless (`lit-api-server/src/dstack/v1/mod.rs`, `get_client_key`):

```
secret = keccak256( dstack.get_key("v1/<derivationPath>", "client") )
wallet_address = secp256k1_pubkey(secret) -> EVM address
```

There is no per-account salt in the derivation. The path is the *only* input. Nothing about the resulting key is bound to the account that created it.

At signing/encryption time the node resolves the path from **the caller's own account registry** (`lit-api-server/src/core/mod.rs:8`, `pkp_id_to_derviation_path`):

```
derivation_u256 = accounts::get_wallet_derivation(api_key, wallet_address)  // reads caller's account
derivation_path = u256_to_derviation_path(derivation_u256)
secret          = get_client_key(derivation_path)                            // re-derives the key
```

[...]

PR #450, opened 2026-06-08
feat(auto-topup): phase 2 — shared lit-billing-auth crate + lit-payments wiring

DashKash54:
**Stack:** PR 2 of 8. Base: `auto-topup-phase-1`. Diff shows only Phase 2.

## What
Extracts the BillingAuth Rocket guard into a new `lit-billing-auth/` crate so both lit-api-server and lit-payments authenticate with identical headers (EIP-712 wallet signature OR API key). Per-service plumbing is supplied via an `AuthResolver` trait pulled from Rocket state.

* **lit-billing-auth** — new crate. BillingAuth enum, request guard with full CPL-285/CPL-286 hardening (rejects precomputed-hash shapes in API-key headers), AuthResolver trait, `is_precomputed_hash_shape` helper. Optional `openapi` feature for OpenApiFromRequest impl.
* **lit-api-server** — deletes `core/v1/guards/billing_auth.rs` (the new crate is the source of truth). `auth_resolver::LocalAuthResolver` bridges the trait to existing in-process eip712.rs + on-chain `get_billing_wallet_address`. Two new internal endpoints for lit-payments to delegate auth verification: `POST /internal/verify_wallet_auth` and `POST /internal/resolve_api_key`.
* **lit-payments** — `auth_resolver::HttpAuthResolver` forwards verification to lit-api-server over the existing X-Internal-Secret channel. Throwaway `GET /_authping` for end-to-end gate testing.

Architectural deviation from plan §17 (option B): on-chain resolver stays in lit-api-server; lit-payments delegates via internal endpoints. Same end-user behavior; avoids ~800 lines of TEE-specific code into the lean side.

## Tests
* 18 unit + integration tests with a mock AuthResolver covering the guard decision tree (valid wallet sig, BadCredentials, Transient, malformed-header fallthrough, valid API key both Bearer and X-Api-Key, CPL-285 precomputed-hash rejection in both header positions, empty key, no headers, missing resolver fails closed).
* 7 lit-api-server integration tests on verify_wallet_auth / resolve_api_key.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #576, opened 2026-07-14
feat(lit-api-server): unsigned `POST /prepare_wallet` derived-address endpoint

clawdbot-glitch003:
## What

Adds `POST /core/v1/prepare_wallet` — the no-signature equivalent of `create_wallet_with_signature`. It returns a freshly derived `{ wallet_address, derivation_path }` with **no EIP-712 signature and no API key**, so a ChainSecured client can obtain a PKP address *before* registering it on-chain and fold the whole owner ceremony into a **single** signed bind UserOp.

Addresses the Chipotle "derived-wallet-address endpoint" side of [flows#532](https://github.com/LIT-Protocol/flows/issues/532):
- New EVM wallet: **3 → 1** passkey prompts
- First exchange connect: **2 → 1**

## Why this is safe (and why it's just a new endpoint)

The existing `*_with_signature` "mint" endpoints don't mint anything on-chain — they generate a random derivation path, derive the key deterministically from the TEE/dstack KDF, and return the address. The client does the real authorization step itself: an on-chain `registerWalletDerivation` inside its owner-signed UserOp. The EIP-712 signature authorizes nothing durable — it only gates the mint API's shape and is trivially self-signed. Dropping it removes UX friction, not a security check. An un-registered response is equivalent to a discarded keypair (compute-only cost).

The `xpub`/client-side-derivation alternative was ruled out: the current derivation is hardened-KDF (no EC structure to retrofit), and Chipotle exports derived secrets as usage API keys, so a published xpub + one leaked child key would compromise the whole tree.

## Design choices

- **Server-generated path only** (not client-supplied). Guarantees 256-bit CSPRNG entropy and sidesteps the cross-account path-collision vector in the security issue below.
- **Unauthenticated**, matching its `_with_signature` siblings — the ChainSecured ceremony has no API key at this point.
- **`CpuAvailable` load-shedding guard** (same one `lit_action` uses): since it's unauthenticated and drives a dstack KDF call per request, this bounds an anonymous caller under CPU saturation.
- **NOT idempotent** — documented prominently: every call mints a new wallet, retries return a different address, concurrent callers each get their own wallet with no server-side dedup. There is no *collision* risk (independent 256-bit paths), only non-idempotency.

## Changes

- `prepare_wallet()` handler + `PrepareWalletResponse` model, route + OpenAPI registration.
- e2e test (`e2e/tests/api/prepare-wallet.spec.ts`) + client method: asserts well-formed address/path and non-idempotency.
- Docs: new `POST /prepare_wallet` section in `docs/management/api_direct.mdx` with the concurrency `Warning`.
- Plan: `plans/prepare-wallet-endpoint.md`.

## Codex adversarial review

Ran `codex challenge` (high). 3 findings, all addressed:
1. **[P1] Unauthenticated unthrottled KDF trigger (DoS)** → added the `CpuAvailable` guard.
2. **[P1] On-chain registration isn't a robust auth boundary (cross-account path hijack)** → this is pre-existing and independent of this endpoint (paths/addresses are already public); corrected the code comment to stop implying it's airtight and cross-referenced the issue. Filed separately as **LIT-Protocol/chipotle#575**.
3. **[P2] Docs overclaimed "no collision / clean revert"** → reworded to "cryptographically negligible" and clarified the revert is per-account.

## Testing

- `cargo check -p lit-api-server` passes.
- Playwright e2e not run here (needs a live server) — please run against a dev instance before merge.

## Follow-ups (out of scope)

- Client-supplied-path lookup (Option B) for the #450 recovery flow — land with the #575 contract fix.
- Deprecate `create_wallet_with_signature` / `add_usage_api_key_with_signature` once flows migrates.


docs/management/api_direct.mdx
```diff
@@ -380,6 +380,10 @@ No API key required. `primaryType` must be `CreateWallet`.
 
 Pass `derivation_path` verbatim into `registerWalletDerivation(adminHash, wallet_address, derivation_path, name, description)` on the AccountConfig contract — until that lands, the PKP exists in MPC but is not registered to any account.
 
+<Note>
```
GTC6244, line 383: Do we want any additional documentation around "why" you might do this ? 

It's very much a `flows` thing ATM ...

GTC6244 (approved):

docs/management/api_direct.mdx
glitch003: yeah i mean the real benefit of it is just streamlining when working in ChainSecure mode.  like instead of making 3 txns and waiting in between them, you just make 1.  i'll update this section with more info like this.  but i think the main bullet points are: 

* Faster
* 1 signature instead of 2 or 3

clawdbot-glitch003: Updated in dcfb45f3 — the Note now spells out the why: one wallet prompt instead of 2–3, faster (no waiting between sequential prompts/txns), same security since `registerWalletDerivation` is the real auth boundary, and calls out that this is how Flows creates ChainSecured wallets today.

PR #577, opened 2026-07-15
fix(contracts): global first-owner binding blocks cross-account PKP hijack (#575)

### LINKED ISSUE #575: Security: public derivation paths + per-account-only uniqueness let any account hijack another account's PKP wallet

clawdbot-glitch003:
Fixes #575.

## What

`registerWalletDerivation` only enforced pkpId uniqueness **within a single account**, but the key is a stateless function of the publicly visible `derivationPath` — so any ChainSecured account could register a victim's `(pkpId, derivationPath)` under its own account and drive the node's signing/encryption endpoints with the victim's key.

- **`AppStorage`**: new `pkpIdToOwnerMaster` mapping (appended to the storage struct — diamond-storage safe).
- **`registerWalletDerivation`**: first master account to register a pkpId owns it forever; cross-account registration reverts with `PKP owned by another account`. The original owner can still re-register (recovery flow #450, re-add after delete).
- **Deletion semantics: Keep it.** The binding deliberately survives `removeWalletDerivation`, closing the delete/re-register race.
- **`backfillPkpOwners`** (owner/config-operator only): one-time migration entry point for wallets registered before this change. Skip-if-set, so it's idempotent and can never re-assign ownership.
- **`getPkpOwnerMaster`** view for auditing the binding/backfill.
- **`tasks/backfill-pkp-owners.ts`**: hardhat task that scans `WalletDerivationRegistered` events, takes the first registrant per pkpId, flags pkpIds registered by multiple masters (possible pre-fix hijacks), submits batches, and verifies. Dry-run by default.

## Deploy order (protects existing users)

1. Cut the new facets in via Safe (`make propose_update_base ...`).
2. **Immediately** run the backfill — until it lands, pre-existing pkpIds have no owner entry and remain claimable:
   `npx hardhat backfill-pkp-owners --diamond <addr> --network base --execute`
3. Re-run until it reports nothing left to bind (covers registrations that raced the upgrade); investigate any multi-master conflicts it flags.

## Testing

- 3 new Foundry tests: cross-account hijack reverts; binding survives hard delete (attacker blocked, owner can re-register); backfill is auth-gated, idempotent, and blocks post-backfill hijack. `forge test`: 49/49 pass.
- Regenerated alloy bindings (forge 1.7.1); `cargo check` clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,14 @@ doesn't describe endpoints the released server lacks.
 - `max_get_keys_count` is now enforced in the key handlers; oversized
   `get_keys` requests are rejected.
 
+### Security
+- `registerWalletDerivation` now enforces a global first-owner binding
+  (`pkpId → master account`): a wallet address can only ever be registered —
+  or re-registered after deletion — by the account that first registered it,
+  closing a cross-account PKP hijack via publicly visible derivation paths
+  (#575). Requires a one-time on-chain backfill for wallets registered before
+  the upgrade (`backfillPkpOwners`).
+
 ## v1.1.10 — 2026-06-22
 
 ### Changed
--- a/lit-api-server/blockchain/lit_node_express/contracts/AccountConfigFacets/AppStorage.sol
+++ b/lit-api-server/blockchain/lit_node_express/contracts/AccountConfigFacets/AppStorage.sol
@@ -119,6 +119,12 @@ library AppStorage {
         EnumerableSet.StringSet nodeConfigurationKeys;
         mapping(string => string) nodeConfigurationValues;
         uint256 serverTriggerValue;
+        // pkpId (wallet address) => master apiKeyHash that first registered it.
+        // Global first-owner binding: prevents a different account from registering
+        // an already-registered pkpId/derivationPath (key derivation is a stateless
+        // function of the public path). Deliberately NOT cleared by
+        // removeWalletDerivation — only the first owner may ever (re-)register.
+        mapping(address => uint256) pkpIdToOwnerMaster;
     }
 
     function getStorage()
--- a/lit-api-server/blockchain/lit_node_express/contracts/AccountConfigFacets/ViewsFacet.sol
+++ b/lit-api-server/blockchain/lit_node_express/contracts/AccountConfigFacets/ViewsFacet.sol
@@ -160,7 +160,27 @@ contract ViewsFacet {
         address walletAddress
     ) public view returns (uint256) {
         AppStorage.Account storage account = getReadOnlyAccount(apiKeyHash);
-        return account.pkpData[walletAddress].id;
+        uint256 derivation = account.pkpData[walletAddress].id;
+        if (derivation == 0) {
+            return 0;
+        }
+        // Cross-account hijack defense (#575). This view is the exact read the
+        // node uses to resolve a derivation path before signing/decrypting, so
+        // enforce the global first-owner binding HERE, not just at registration:
+        // if the wallet has an owner and the resolving (master) account is not
+        // that owner, the local pkpData entry is a stale pre-fix hijack
+        // registration — fail closed instead of leaking the victim's path.
+        // owner == 0 means a pre-migration wallet not yet backfilled: fall
+        // through so signing keeps working until backfillPkpOwners runs.
+        AppStorage.AccountConfigStorage storage s = AppStorage.getStorage();
+        uint256 owner = s.pkpIdToOwnerMaster[walletAddress];
+        if (owner != 0) {
+            uint256 resolvedMaster = s.allApiKeyHashesToMaster[apiKeyHash];
+            if (owner != resolvedMaster) {
+                revert AppStorage.InvalidRequest("PKP owned by another account");
+            }
+        }
+        return derivation;
     }
 
     /// @notice Return the master apiKeyHash that first registered a pkpId, or 0 if
--- a/lit-api-server/blockchain/lit_node_express/contracts/AccountConfigFacets/ViewsFacet.sol
+++ b/lit-api-server/blockchain/lit_node_express/contracts/AccountConfigFacets/ViewsFacet.sol
@@ -163,6 +163,15 @@ contract ViewsFacet {
         return account.pkpData[walletAddress].id;
     }
 
+    /// @notice Return the master apiKeyHash that first registered a pkpId, or 0 if
+    ///         the pkpId has never been bound (pre-migration wallet or never registered).
+    /// @dev The binding survives removeWalletDerivation by design; used to audit the
+    ///      global first-owner rule and the one-time backfill migration.
+    function getPkpOwnerMaster(address pkpId) public view returns (uint256) {
+        AppStorage.AccountConfigStorage storage s = AppStorage.getStorage();
+        return s.pkpIdToOwnerMaster[pkpId];
+    }
+
     function listApiKeys(
         uint256 accountApiKeyHash,
         uint256 pageNumber,
--- a/lit-api-server/blockchain/lit_node_express/contracts/AccountConfigFacets/WritesFacet.sol
+++ b/lit-api-server/blockchain/lit_node_express/contracts/AccountConfigFacets/WritesFacet.sol
@@ -68,6 +68,10 @@ contract WritesFacet {
         uint256 indexed apiKeyHash,
         address indexed pkpId
     );
+    event PkpOwnerBackfilled(
+        address indexed pkpId,
+        uint256 indexed masterHash
+    );
     event UsageApiKeyRemoved(
         uint256 indexed accountApiKeyHash,
         uint256 indexed usageApiKeyHash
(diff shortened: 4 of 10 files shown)


GTC6244 (approved):

clawdbot-glitch003:
## Codex adversarial review — findings + fixes

Ran an independent adversarial pass (OpenAI Codex, high reasoning) against the diff. It found one real design gap that I'd under-weighted, plus two task-correctness issues. All addressed in the latest commit.

**[P1 — fixed] Pre-existing hijacks survived the fix.** The registration-time binding stops *new* cross-account registrations, but the node resolves derivation paths through `getWalletDerivation(callerHash, wallet)` (`core/mod.rs:8` → `accounts::get_wallet_derivation` → the contract view), which reads the caller's *account-local* `pkpData` and never consulted `pkpIdToOwnerMaster`. So a hijack registered **before** the upgrade leaves a stale local entry the node would still honor — deploy + backfill alone wouldn't rescue an already-compromised wallet.
Fix: `getWalletDerivation` now fails closed — if the wallet has an owner binding and the resolving master isn't that owner, it reverts. `owner == 0` (not yet backfilled) falls through so signing keeps working during the migration window. Because this is the exact read the signing path uses, it neutralizes already-happened hijacks on-chain, with no Rust change and no ABI change. This folds the issue's node-side defense-in-depth item into the contract instead of leaving it as a follow-up.

**[P1 → P2 — fixed] Task only logged conflicts and proceeded.** Multi-master pkpIds (evidence the attack already happened) were printed but the run continued. Now they're a hard stop under `--execute` unless the operator reviews and passes `--allow-conflicts`, with a note that the later registrant's stale row still needs separate removal.

**[P2 — fixed] Event ordering assumed.** The scan trusted provider log order. Now it collects all logs and sorts explicitly by `(blockNumber, transactionIndex, logIndex)` so "first registrant wins" is actually enforced.

**[P2 — fixed] No finality guard.** The scan used `latest`; a reorg could rerank the first registrant. Now it scans to the `finalized` block tag (or `--confirmations N` behind head) and fixes the upper bound up front.

New tests: pre-existing-hijack read fails closed; `owner == 0` legacy read still resolves; a storage-slot sanity assertion guards the `vm.store` layout used to plant the pre-fix state. `forge test`: 51/51 pass.

One residual (documented, not auto-fixable): if an attacker registered a wallet *before* the real owner ever did, the backfill binds the attacker as first owner. That's inherent to reconstruction from history — the conflict report flags these for manual review.

Issue #600, opened 2026-07-23
gVisor lane: request-supplied startup_script runs under the bundle CID's identity+key without being content-addressed


clawdbot-glitch003:
## Summary

On the gVisor any-language lane, a request-supplied `startup_script` executes **under the bundle CID's identity and key material** but is itself **not part of the content-addressed, authorized artifact**. Because CID-derived key access (`lit get-action-private-key`, `getLitActionPublicKey({ipfsId})`, group/PKP decrypt) is keyed by the bundle CID alone, anyone holding an execute-scoped usage key for the bundle's group can run **arbitrary code that wields the audited bundle's key** — by sending their own startup script against the cached bundle CID.

This breaks the property the JS lane provides and that our sovereignty story depends on: "the exact audited bytes are what ran, and only that code can hold the CID-derived key." On this lane, the CID no longer pins the code that runs.

## Where it lives

- The endpoint forwards the request's `startup_script` straight through to the runner, beside the bundle, while authorizing only on the derived bundle checksum:
  - `lit-api-server/src/core/core_features.rs` — `lit_binary_action`: auth is `can_execute_action(api_key, cid_hash)` on the **bundle** checksum (`resolve_binary_bundle` derives it from the tar bytes), and `ExecutionOptions { action_ipfs_id: Some(checksum), startup_script: request.startup_script.clone()… }`. The startup script never enters the checksum.
- The supervisor materializes and runs the request script verbatim, request-script-wins:
  - `lit-actions/gvisor-server/src/supervisor.rs` — "the request-supplied script wins (one cached bundle, many scripts)".
- Key derivation is keyed by `ipfs_id` (= the bundle checksum), independent of the script:
  - `lit-api-server/src/actions/client/op_code_helpers/private_keys.rs` — `get_lit_action_key(ipfs_id)` → private key / public key / wallet address.
- The behavior is documented as intentional today:
  - `lit-actions/gvisor-server/README.md`: *"authorization keys on the **bundle** CID alone; the per-request startup script is authenticated by API-key authorization, not content-addressing."*

So the current guarantee is only: *"someone authorized for this bundle ran **some** script with this bundle's key."* It is **not**: *"the audited bundle's own logic ran."*

## Concrete attack

1. Vendor publishes an audited bundle; its CID `Qm…` is registered in a group; the customer encrypts a secret to `getLitActionPublicKey({ipfsId: Qm…})` and authorizes the CID.
[...]

PR #601, opened 2026-07-23
ci: manual workflow to toggle GVISOR_RUNNER_ENABLED on the test diamond

clawdbot-glitch003:
## What

A `workflow_dispatch` job that flips the on-chain **`GVISOR_RUNNER_ENABLED`** gate for the **test/next** environment without a redeploy.

`GVISOR_RUNNER_ENABLED` is the third of the three fail-closed axes gating `POST /lit_binary_action` (CPL-359/361). On the test env the other two are already on — the `gvisor` cargo feature is built, and `LIT_GVISOR_ENABLED` is substituted `true` by `deploy-staging.yml` — but the on-chain key sits **unset** on the test diamond, so every binary-action call 503s with *"disabled network-wide by contract configuration."* This job is the last switch.

## How it's authorized

It signs `WritesFacet.setNodeConfiguration("GVISOR_RUNNER_ENABLED", <value>)` with `secrets.PHALA_DSTACKAPP_PRIVATE_KEY` — the same secret `manual_contract-upgrade.yml` uses as the `DEPLOYER_KEY` for owner-only `diamondCut` upgrades on Base. `setNodeConfiguration`'s guard is `revertIfNotApiPayerOrOwner`, which accepts the diamond owner, so the deployer key clears it. The first run logs `cast wallet address` before spending gas, so the signer is visible immediately.

## Safety

- The **test diamond** `0x98e501fab2d60a5119a185e1563f10cb54bc6068` is hardcoded in `env:`. The job cannot touch the prod diamond (`0xaaaaa9120fe271f653cfdb6bf400db93d2dea7aa`); prod's gVisor gate stays off behind the governed contract-upgrade path.
- Reads back the value before and after via `nodeConfigurationValue(string)`.
- No new secrets; no GitHub Environment gating involved.

## Why

Unblocks external validation of the gVisor any-language lane end-to-end (a client-encrypted-secret-in, CID-key-signed-receipt-out workload running an off-the-shelf agent harness in the sandbox). Related: the lane's one structural caveat is filed as #600.

Nodes re-read the `ChainConfig` snapshot within ~30s of the tx landing.

### DIFF
--- a/.github/workflows/manual_gvisor-toggle-test.yml
+++ b/.github/workflows/manual_gvisor-toggle-test.yml
@@ -1,0 +1,58 @@
+name: "Manual: toggle GVISOR_RUNNER_ENABLED (test diamond only)"
+
+# Flip the on-chain gVisor gate (CPL-361) for the TEST/next environment
+# without a redeploy. GVISOR_RUNNER_ENABLED is one of three fail-closed axes
+# on POST /lit_binary_action; the other two (the `gvisor` cargo feature and
+# the LIT_GVISOR_ENABLED env) are already ON for staging/test deploys, so
+# this key is the last switch. Nodes re-read the ChainConfig snapshot within
+# ~30s of the tx landing.
+#
+# Signs with PHALA_DSTACKAPP_PRIVATE_KEY — the same secret manual_contract-
+# upgrade.yml uses to sign owner-only diamondCut upgrades on Base, i.e. the
+# diamond owner, which WritesFacet.setNodeConfiguration's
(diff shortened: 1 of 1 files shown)


glitch003:
closing - going to redeploy

clawdbot-glitch003:
FYI from the Ember side: we've pivoted our PoC away from the gVisor any-language *workload* lane toward per-client Phala CVMs running per-vendor Docker images (RAG needs persistent storage + long-running jobs that the serverless isolate model fights), so we no longer need this toggle as a blocker. No action needed on our account. Issue #600 stands as a valid finding regardless of direction.

PR #627, opened 2026-08-18
docs(plans): PRD for Lit Chat, a private chat app served from the TEE

GTC6244:
## Summary

PRD for **Lit Chat** — a Claude/ChatGPT-style web app whose serving stack runs in a TEE and whose chat history is stored outside the TEE as ciphertext under keys that only exist inside the enclave. Covers the three parts of the ask: the chat web app, inference in the TEE, and encrypted off-TEE storage.

Three decisions carry the design:

**lit-chat gets its own Phala CVM.** Not a Lit Action: the action path is request/response with a 1 MB buffered response (`lit-api-server/src/actions/client/mod.rs:27`), no persistence ops, $0.01/sec billing (`stripe.rs:33`, so a 20s generation costs $0.20 before model cost), and billing-flush failures cancel actions mid-flight. Not a sibling container in the core CVM either: any container on `/var/run/dstack.sock` can call `get_key` with arbitrary paths and derive every PKP on the platform, `dstack-ingress` binds 443 to a single `TARGET_ENDPOINT`, and every compose change is a Safe-multisig ceremony a consumer app can't iterate through. Lit Actions come back in P2 as the *tool* layer, which is where they're strong.

**Storage is ciphertext-only, off-TEE.** Per-user KEK from `get_key`, per-conversation DEK, AES-256-GCM extended with AAD binding each row to `(conversation_id, message_id, seq, role)` so a DB operator can't swap, reorder, or re-label rows. First implementation of the CPL-364 pattern. The doc is explicit about what this does *not* give: no tamper-evidence over time, so tail truncation and wholesale rollback are disclosed rather than hand-waved.

**Sessions and magic-link tokens are signed in-enclave, never DB-rooted.** If the database decided whose account a token unlocks, an operator with write access could mint a session for any user and have the enclave decrypt their history — collapsing the headline claim to a policy promise. Account identity derives straight from the email (`HKDF(ns, lower(email))`), so the schema holds no email, no hash of one, and no UUID mapped to one.

Also included: a cost-plus pricing model on the existing Stripe ledger (§8), a threat model that names the residuals (§6), claim-discipline rules for marketing copy (§7.2, §8.6), and a Venice.ai comparison (§12).

## Pricing (§8)

Cost-plus passthrough on OpenRouter's reported per-generation cost, metered against the `lit-billing-core` Stripe credit ledger — chat becomes a third consumer of that ledger, not a second billing system. Markup written in as **5%, explicitly a placeholder**. Three things the arithmetic surfaced:

- The ledger is integer cents and bottoms out at $0.01; a chat message costs a fraction of that, so per-message cent-granular charging over-bills by 10–100×. Needs a µUSD accumulator that flushes floor-to-cents. This is the one billing piece that can't be reused.
- Stripe's ~2.9% + $0.30 on the $5.00 minimum top-up is ~8.9% of the top-up — the markup doesn't cover the cost of collecting the money.
- The 25% LITKEY discount applies at *funding* (`lit-payments/src/rate.rs`), so it discounts whatever the credits later buy. Cost + 5% funded with LITKEY sells $1.00 of OpenRouter cost for ~$0.79. Needs a decision before any LITKEY messaging mentions chat.

## Review

Adversarial review run against the doc during drafting. Four findings, all fixed before this PR:

| # | Finding | Fix |
|---|---|---|
| 1 | "Post-quantum encrypted" omitted the KMS key-provisioning channel — the app key the KEKs derive from is released to the CVM over an attested ECDH channel, so a recorded handshake plus the DB ciphertext is a harvest-now-decrypt-later path to the archive | Added to the §7.2 exception list with the scope and the caveat that closing it needs PQ-hybrid transport upstream in dstack |
| 2 | `magic_links` reintroduced the DB-rooted auth path §5.1 exists to close | Token carries signed identity, row is a replay guard only |
| 3 | No way to resolve a returning account holder to their history | Derivation *is* the lookup — no email-derived column at all |
| 4 | `deleted_at` contradicted the stated immediate-deletion guarantee | Hard delete, tombstone removed |

## Test plan

- [x] Docs-only change (one new file under `plans/`, matching the precedent set by #551 and #561). No code paths touched, so no test suite, coverage audit, or eval run applies.
- [x] Every repo claim in the doc verified against source: action response cap, per-second billing, `get_key` arbitrary-path derivation and keccak wrap, single `TARGET_ENDPOINT` on dstack-ingress, `aes.rs` construction, CPL-367 rate limiter, LITKEY discount mechanics, `MIN_TOPUP_CENTS`, `STARTER_CREDITS_CENTS`, `verify-attestation.yml`, `lit-static/dapps/verify`.
- [x] All internal cross-references resolve.
- [x] No VERSION bump or CHANGELOG entry — repo convention is that plan docs don't carry either.

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -124,27 +124,51 @@ yanked = "warn"
 #   tar 0.4.45 -> 0.4.46 was attempted but BLOCKED: deno_npm_cache requires
 #   =0.4.45, so GHSA-3pv8 (tar PAX desync) stays Dependabot-dismissed until deno
 #   bumps. No RUSTSEC id assigned, so no ignore entry needed here.
+# - 2026-08-19: RUSTSEC-2026-0258 (h2 unbounded empty DATA frames) published and
+#   failed all six h2-bearing workspaces at once. Two h2 lines were in play:
+#   * h2 0.4.12/0.4.13/0.4.14 -> 0.4.16 (the patched release) in lit-actions,
+#     lit-core, lit-api-server, lit-billing-core, lit-payments and lit-triggers.
+#     Lockfile-only; h2 0.4.16's dep set and MSRV (1.63) are unchanged, so the
+#     bump is a version+checksum edit with no resolver cascade. NB: running
+#     `cargo update -p h2 --precise 0.4.16` under cargo 1.91 ALSO downgrades
+#     socket2 0.6.x -> 0.5.10 and shuffles windows-sys, because the MSRV-aware
+#     resolver re-resolves the whole graph against 1.91 while these lockfiles
+#     were generated with a newer cargo. Bump h2 in isolation and verify with
+#     `cargo metadata --locked` rather than taking the resolver's cascade.
+#   * h2 0.3.27 stays: 0.3.27 is the FINAL 0.3.x and the fix was never
+#     backported, so no lockfile bump exists. Held by rocket 0.5.1 -> rocket_http
+#     -> hyper ^0.14 in all six, plus the zerossl fork's reqwest 0.11.27 in
+#     lit-actions/lit-core/lit-api-server. Entry added below.
+#   The same CI run also failed lit-actions on RUSTSEC-2026-0249 (smartstring
+#   unmaintained; repo archived 2026-05-03, all versions affected, no successor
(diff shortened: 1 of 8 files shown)


GTC6244:
moved to ember

PR #629, opened 2026-08-20
fix(deps): patch h2 for RUSTSEC-2026-0258, record the two unfixable blockers

GTC6244:
## Summary

Rust CI went red on **every** matrix job, on main and every open branch. Nothing in the repo changed — two advisories were published to the RustSec database on 2026-08-18 and `cargo deny` picks the DB up fresh at runtime, by design ("Advisories published AFTER this snapshot will NOT be in this list, so they will fail CI — that is the point").

| Advisory | What | Jobs hit | Fixable? |
|---|---|---|---|
| RUSTSEC-2026-0258 | h2 accepts and queues empty DATA frames without limit → unbounded memory / overflow panic. Low severity, patched in 0.4.16 | all 5 | Partly — see below |
| RUSTSEC-2026-0249 | smartstring unmaintained; repo archived 2026-05-03, all versions affected, no successor | lit-actions | No |

## h2: one line patched, one line blocked

Two h2 versions are in the graph, and only one of them can move.

**Patched** — `hyper 1.x → h2 0.4.12 / 0.4.13 / 0.4.14` bumped to **0.4.16** in all six h2-bearing lockfiles. Lockfile-only: 0.4.16's dependency set and `rust-version` (1.63) are byte-identical in effect to 0.4.14, so this is a version + checksum edit with no graph change. Reached via reqwest / alloy / tonic.

**Blocked** — `rocket 0.5.1 → rocket_http → hyper ^0.14 → h2 0.3.27`. **0.3.27 is the final 0.3.x release and the fix was never backported**, so there is no version to bump to. Also reached through the zerossl fork's `reqwest 0.11.27` in lit-actions / lit-core / lit-api-server. That is the same pair of blockers already documented for RUSTSEC-2025-0134 (rustls-pemfile), so this entry joins the existing rocket + zerossl group rather than describing a new problem.

`smartstring` is the deno pin again: sole path is `swc_ecma_lexer 26.0.0 ← deno_ast =0.53.2`, and it appears in no other lockfile. Both remainders get `ignore` entries with the file's usual BLOCKER / WAITING ON justification, and the blocker-summary counts are updated so the file stays internally consistent (deno 9 → 10, rocket 4 → 5, no-upstream-fix 4 → 6).

## A trap worth knowing about

The obvious command does the wrong thing here. `cargo update -p h2 --precise 0.4.16` under cargo 1.91 **also** downgrades `socket2 0.6.x → 0.5.10` and shuffles `windows-sys` and `syn` (2.0 → 1.0 in one workspace). The MSRV-aware resolver re-resolves the whole graph against 1.91, while these lockfiles were written by a newer cargo — so it "fixes" h2 and quietly rewrites a dozen unrelated pins that CI would then build with under `--locked`. Setting `CARGO_RESOLVER_INCOMPATIBLE_RUST_VERSIONS=allow` makes it worse, not better.

So the h2 bump here is a surgical single-package edit, and the trap is written into the deny.toml history for whoever does the next one.

## Test plan

- [x] `cargo deny --all-features check --config deny.toml -W advisory-not-detected -W unmatched-source advisories sources` — **all six CI matrix workspaces report `advisories ok, sources ok`** (lit-actions, lit-core, lit-api-server, lit-api-server/blockchain/rust_generator_and_deployer, lit-billing-core, lit-payments).
- [x] Every touched lockfile accepted under `cargo +1.91 metadata --locked` — the same flag `clippy` / `build` / `test` use, so the lock is self-consistent and cargo does not want to change it.
- [x] `cargo +1.91 build --locked -p h2@0.4.16` compiles clean.
- [x] h2 0.4.16 checksum verified independently against the downloaded `.crate` (`shasum -a 256`), not just taken from cargo's output.
- [x] Lockfile diffs contain h2 lines only — verified mechanically, no incidental resolver churn.
- [x] No source files touched, so `fmt` and `clippy` are unaffected.

## Notes

- `lit-triggers` is not in the CI matrix, so it was not failing, but it carries the same vulnerable h2 0.4.14 — bumped it too rather than leaving one workspace behind on a one-line fix.
- Unrelated and pre-existing: `lit-triggers` fails `cargo deny` on a **ruint** advisory ("Uint shift operations: incorrect overflow flags and truncated shift amounts"), confirmed present before this change. It gates nothing today because the crate is absent from every workflow. Worth a look separately — either fix it or add it to the matrix, because right now it is an unwatched workspace.

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -124,27 +124,53 @@ yanked = "warn"
 #   tar 0.4.45 -> 0.4.46 was attempted but BLOCKED: deno_npm_cache requires
 #   =0.4.45, so GHSA-3pv8 (tar PAX desync) stays Dependabot-dismissed until deno
 #   bumps. No RUSTSEC id assigned, so no ignore entry needed here.
+# - 2026-08-19: two advisories published on 2026-08-18 turned every Rust CI job
+#   red at once (all six matrix workspaces). Neither is fixable in-house:
+#   * RUSTSEC-2026-0258 (h2 unbounded empty DATA frames, low severity). Two h2
+#     lines were in the graph. The hyper-1.x line was bumped 0.4.12/0.4.13/
+#     0.4.14 -> 0.4.16 (the patched release) in lit-actions, lit-core,
+#     lit-api-server, lit-billing-core, lit-payments and lit-triggers —
+#     lockfile-only, and h2 0.4.16's dep set and MSRV (1.63) are identical to
+#     0.4.14, so it is a version+checksum edit with no graph change. h2 0.3.27
+#     stays: it is the FINAL 0.3.x and got no backport, so no bump exists. Held
+#     by rocket 0.5.1 -> rocket_http -> hyper ^0.14 everywhere, plus the zerossl
+#     fork's reqwest 0.11.27 in lit-actions/lit-core/lit-api-server.
+#     TRAP for whoever does the next bump: `cargo update -p h2 --precise 0.4.16`
+#     under cargo 1.91 also downgrades socket2 0.6.x -> 0.5.10 and shuffles
+#     windows-sys/syn, because the MSRV-aware resolver re-resolves the whole
+#     graph against 1.91 while these lockfiles were written by a newer cargo.
+#     Bump the one package in isolation and prove it with `cargo metadata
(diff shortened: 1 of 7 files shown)


GTC6244:
Close - this is now fixed.

PR #634, opened 2026-09-10
fix(accounts): prevent permanent signer wedge after RPC outage

clawdbot-glitch003:
## Incident

On 2026-09-03 ~15:28 UTC, the Base RPC (Alchemy) returned a storm of HTTP 403s to prod. Two signer-pool payers (`0xC5b6…3745`, `0x9Fde…53BE`) were mid-transaction; their in-memory nonce state was poisoned, and neither landed a transaction again. From then until 2026-09-09, GCP logs show those two wallets absorbed **71 of the last 73 lease grants**, and each borrowed lease was force-freed as stale (~10-15s held) while the borrowing HTTP request hung forever — most visibly `POST /core/v1/new_account` timing out with 0 bytes.

## Failure chain and fixes

Each fix independently breaks a link:

1. **Poisoned nonce cache → transactions that can never mine.** Alloy's `NonceFiller` (`SimpleNonceManager`) increments its cached nonce optimistically and never rolls it back after a failed/dropped broadcast, so one RPC outage mid-send leaves the signer emitting nonce-gapped (queued-forever) transactions.
   **Fix:** sends now pin nonces from a locally managed per-signer `NonceCache`. A cache hit costs no extra RPC (same as `NonceFiller` on the happy path); the difference is that **any broadcast failure or receipt timeout invalidates the entry**, so the next send re-derives the nonce from the chain and the signer self-heals. `advance()` uses `max()` so an out-of-order write from the pool's rare shared-lease fallback can't rewind the counter (mined nonces are chain truth, so `max` can never poison it). A rare nonce collision surfaces as "nonce too low" and is handled by the existing single retry.

2. **Unbounded `get_receipt()` → requests hang forever.** A transaction that never mines pinned the HTTP request (and its signer lease) indefinitely.
   **Fix:** all receipt waits are bounded at 30s (`TX_RECEIPT_TIMEOUT`; chains here mine ~2s blocks). This includes the rebalance funding transactions, which run inside the pool task's select loop and would otherwise stall **every** lease grant.

3. **Stale-freed signers monopolize the lease queue.** `release_stale_leases` freed the lease but didn't touch `last_request`, and `run_pool` grants the oldest-`last_request` idle signer first — so a wedged signer permanently sorted to the front and soaked up nearly all traffic.
   **Fix:** a force-freed signer now rotates to the back of the LRU order.

## Perf: 2s receipt polling

Alloy defaults the receipt poller to **7s** for non-local HTTP transports (`alloy-rpc-client/src/client.rs`), but our chains mine ~2s blocks. Both providers now set a 2s poll interval (`RPC_POLL_INTERVAL`), so writes confirm ~2-5s sooner and signer leases are held correspondingly shorter.

## Testing

- Unit tests for stale-lease rotation (`signer_pool::tests`) and the `NonceCache` semantics (advance/no-regress/invalidate); full `cargo test --lib` passes (344 tests).
- `cargo clippy --all-targets` clean.


clawdbot-glitch003:
Hardening pass (commit 4de11abb) addressing the blocking findings from the adversarial cross-model review (gpt-6-astra):

- **Finding 2 + 8 (nonce collisions / cache interleavings):** `NonceCache` now *reserves* at read time — `reserve`/`reserve_seeded` hand out a nonce and advance the counter in one locked step, so concurrent borrowers of one signer get distinct nonces. Advance-on-receipt is gone (no more resurrect-a-consumed-nonce interleavings), and the retry path invalidates on failure.
- **Finding 1 (timeout doesn't bound the RPC awaits):** every step of the send pipeline — simulation, nonce fetch, broadcast, receipt — now has a `tokio::time::timeout` hard deadline (`RPC_STEP_TIMEOUT` 10s; receipt outer bound = watcher timeout + 10s grace). Pool housekeeping is bounded at 30s and rebalancing is spawned off the dispatcher task, so a blackholed RPC can no longer freeze lease grants.
- **Finding 6 (cross-borrower release):** leases carry a per-grant id; release is a no-op unless the id matches. `STALE_LEASE_SECS` raised 10s → 150s (above the ~130s worst-case bounded send), so stale cleanup is pure leak recovery instead of a systematic lease-sharing machine.
- **Finding 3 (split-brain with NonceFiller):** rebalance repatriations pin nonces from the same allocator contract writes use; admin sends (funding, `setApiPayers`) pin a fresh pending fetch per send. `NonceFiller` remains installed but inert.
- **Finding 9 (fallback concentration):** the all-busy fallback now takes over lease ownership and refreshes `in_use_since`, so consecutive fallback grants rotate across busy signers.
- **Finding 4 (partial rebalance):** one failed wallet no longer aborts the remaining drains/fundings (full reconciliation tracked separately).

Not addressed here (pre-existing design scope, tracked as follow-ups): idempotency/reconciliation for transactions that mine after a timeout (finding 5), and fee-replacement for stuck pending predecessors (finding 7).

`grant_lease`/`release_lease` extracted as pure functions with unit tests; `NonceCache` tests rewritten for reservation semantics. 349 tests pass, clippy clean.

GTC6244 (approved):

PR #638, opened 2026-09-11
fix(keychain): address dashboard/audit/docs findings from product test (KC-01..13)

clawdbot-glitch003:
## Summary

Fixes all 13 findings from the black-box product test of Lit Agent Keychain (report in `/tmp/keychaintest/bugs.md`, not committed).

### Backend
- **KC-04 audit integrity** — `access_log` now snapshots `secret_name`/`agent_name` at write time and `secret_id` is no longer an `ON DELETE SET NULL` FK. Deleting a secret keeps its history attributable; `/api/audit` returns `secret_deleted: true` for those rows. Migration backfills existing rows.
- **KC-05 setup-token lifecycle** — new `GET /api/setup-tokens` and `DELETE /api/setup-tokens/<hash>`; dashboard section “Setup agents (full access)” lists and revokes them. Authorize page gains a label field and states that tokens don’t expire and aren’t revoked by sign-out.
- **KC-03 duplicate agent names** — `POST /api/agents` returns `409 agent_name_exists` when an active agent already has that name.
- `/api/*` guard failures now return JSON `{error, detail}` instead of Rocket’s HTML (the test hit this with a runtime key on a management endpoint).

### Dashboard
- **KC-01** full-width responsive layout, contained scrolling for code/tables, consistent inputs.
- **KC-03/06** policy editor dialog keyed by immutable agent id (prefilled), with quota + expiry editing and rolling-24h wording.
- **KC-07** one-time key shown in a masked field with Copy/Show and an explicit Done (warns if not copied), plus a Node snippet using the user’s own secret name.
- **KC-08** every section has loading / failed (+Retry) states; one failure no longer blanks the page.
- **KC-09/10** revoke dialog matches docs; confirmation with affected agents before downgrading in-TEE-only → plaintext.
- **KC-11** rotate via a multiline dialog; lazy “Versions” list per secret with recovery instructions.
- **KC-12/13** Cancel discards the draft; actionable validation messages; input preserved on rejected save.
- Getting-started steps and a 30–60 s first-save expectation.

### Docs (KC-02)
Node quickstart uses `npm install @lit-protocol/keychain` (Node’s ESM loader rejects `https:` imports; package is live at 0.1.0), with a curl-download fallback and Deno/browser URL import. Fixed `secrets.grant/reference` → `keychain.*` in `llms-full.txt`. Documented setup-token lifecycle, new audit fields, agent-name uniqueness.

## Testing
- `cargo clippy --all-targets -- -D warnings`, `cargo fmt --check`, `cargo test` (35 passed).
- `node --check static/app.js`.
- Rendered the dashboard against a mock API in headless Chrome at 1440 px and 390 px; document `scrollWidth` at 390 px is 375 (no page-level overflow), the audit table and code blocks scroll within their containers.
- Not exercised against a live Postgres/Chipotle: the migration and the new endpoints need a dev deploy smoke test (migration drops `access_log_secret_id_fkey`, which is the Postgres default constraint name from the foundation migration).


clawdbot-glitch003:
Codex review (gpt-6-astra) ran against this PR: 1 P1, 2 P2. All three addressed in ff79caa8:

- **P1** grant/delete race after removing the `access_log.secret_id` FK → `issue_grant` re-reads the secret `FOR SHARE` inside the quota transaction; missing row → audited 404, and the share lock blocks a concurrent delete until commit.
- **P2** agent-name uniqueness not atomic → partial unique index `(tenant_id, name) WHERE revoked_at IS NULL` (migration renames existing active dupes with an id suffix); on unique violation the just-minted Chipotle key is removed before returning 409.
- **P2** `novalidate` let half-typed quota/expiry serialize as empty → explicit `badInput` checks on both the create form and the policy dialog.

PR #647, opened 2026-09-14
docs: PKP vs derived key — funds live in a PKP, signers use the identity key

clawdbot-glitch003:
## Summary

Follow-up to #644. The existing docs contrasted PKPs and action identity keys mostly by "code evolves vs. code is the identity". The sharper, more useful rule is about **where the funds sit**:

- **The Lit key holds funds or on-chain state → PKP.** The address survives code changes (add the new CID to the group), and group membership gives a governable, changeable answer to "which code may spend from here". No sweep on upgrade.
- **The Lit key is only a signer → derived / identity key.** Proofs and attestations, an owner on a Safe or ERC-4337 account, per-user or per-secret authority. Nothing sits at the address, so "new bytes = new address" costs nothing, and there is no PKP to mint or group to maintain.

## Changes

- **`lit-actions/index`**: added a "survives a code change" row to the key comparison, rewrote the "best for" row around holds-funds vs. is-a-signer, and added a **Which one should hold the funds?** section with the rationale, a situation table, and the common mixed shape (identity key authorizes, PKP or contract executes).
- **`lit-actions/derived-actions`**: new **Signer, not treasury** section covering the three safe shapes (proof signer, smart-contract-wallet signer, authorizer for a PKP-backed executor), plus a new operational rule. Table row reworded from "wallet" to "signer".
- **`lit-actions/patterns`**: warning under *A Wallet Bound to the Action* explaining the demo deposits at the derived address for self-containment and what to do for real balances. Also fixed the stale "Examples §13" link, which pointed at a heading that is now §14.
- **`lit-actions/examples`**: short caveats on §14 (action-bound wallet) and §17 (Solana signer) framing them as signer demonstrations rather than custody recommendations.

No example code changed. The runnable `action-bound-wallet` and `solana-signer` examples are unchanged; the docs now explain how to read them.

## Verification

- All new and changed internal anchors resolve (`#which-one-should-hold-the-funds`, `#signer-not-treasury`, `#14-a-unique-immutable-wallet-per-user-bound-to-the-action`).
- Edited MDX scanned for stray `{` / `<` outside code.
- Mintlify preview build not run (CLI unavailable here).


glitch003:
i decided i don't like these changes and i'm going to keep the existing docs

PR #654, opened 2026-09-15
hotfix(v1.1.10): backport tx-latency, signer-wedge, read-after-write and Stripe-search fixes (#556 #634 #646 #650)

clawdbot-glitch003:
## Why

Prod runs `release/v1.1.10` (tip `ae36a42c`), which is ~2 months behind `main`. Four merged fixes that directly affect chain-secured management writes and billing are not on prod:

| PR | What it fixes | Prod symptom today |
|---|---|---|
| #634 | Receipt poller ticks every **2s instead of alloy's 7s default**; hard deadlines on every RPC step; nonce allocator + lease ids so payers can't wedge | Every `add_group` / `add_usage_api_key` / `new_account` waits a full extra poll cycle (~7s each). Root cause of the 2026-09-03 signer-wedge incident. |
| #646 | `new_account` waits for the account to be visible on the read RPC before `registerWalletDerivation` | Intermittent `NoAccountAccess` 500 on account creation |
| #650 | `add_usage_api_key` / `add_group` wait for the write to be visible before returning | A client that immediately uses a fresh usage key gets a spurious 401 "does not resolve to any account"; immediate group mutation reverts `GroupDoesNotExist` |
| #556 | Billing guard no longer caches a duplicate zero-credit Stripe customer when Stripe Search lags (issue #555) | 402 on first action for newly funded accounts; duplicate customers |

Concrete motivation: Lit Agent Keychain's first sign-in issues three sequential chain-secured writes (`add_group` ×2, `add_usage_api_key`) and then immediately executes with the new key. On prod that is ~30s of receipt polling plus a race into the #650 401. With this backport each write confirms in ~2-4s and the follow-up execute is safe.

## Cherry-pick notes

All four commits are `cherry-pick -x` of the `main` merges, in merge order. Deviations from `main`:

- **#634:** dropped the hunk touching `remove_wallet_derivation` — that function comes from #553 (dashboard PKP delete), which is not on this line.
- **#646:** took the code and CHANGELOG only. Its `Cargo.lock` rustls 0.23.45 bump and `deny.toml` RUSTSEC-2026-0285 ignore are **not** included; lockfiles are untouched from `release/v1.1.10` (`--locked` builds unchanged).
- **CHANGELOG:** the `Unreleased → Fixed` bullets from #634/#646 were merged into this branch's existing `Unreleased` section.

## Verification (local, this branch)

```
lit-api-server:   cargo test --locked --all-features --lib   264 passed, 0 failed
lit-billing-core: cargo test --locked --all-features          74 passed, 0 failed
both:             cargo fmt --check, cargo clippy --locked --all-features -D warnings   clean
```

Not runnable locally: k6 correctness/staging soak.

## Known CI caveat

`rust-ci`'s **cargo-deny advisories** step already fails on the untouched `release/v1.1.10` baseline (RUSTSEC-2026-0190/0204/0220/0221/0253/0258, …) because the lockfiles predate `main`'s advisory cleanups. This PR does not change lockfiles, so expect that job red for reasons unrelated to the diff. Fixing it means bumping lockfiles on the release line, which I kept out of a hotfix.

## Releasing

After merge, tag the branch tip (e.g. `v1.1.11`) to trigger `deploy-prod-1-propose`. Note that workflow's `verify-staging-passed` gate looks for a green Deploy Staging soak status on the tagged commit's first-parent ancestors; a release-branch commit may not have one, in which case use the manual dispatch with `skip_staging_check=true`.


glitch003:
everything already in main, closing

PR #659, opened 2026-09-15
ci: deploy lit-payments and lit-triggers to Railway

GTC6244:
Adds GitHub Actions workflows to deploy `lit-payments` and `lit-triggers` to Railway via the Railway CLI, mirroring the `lit-static` release flow. Pushes to `main` deploy to the Railway **staging** environment and `v*` tag pushes deploy to **production**, with a manual `workflow_dispatch` fallback. Each `railway up` uses the build context its Dockerfile needs (repo root for lit-payments so it can reach `lit-billing-core`; `lit-triggers/` for lit-triggers) and authenticates with an environment-scoped Railway project token, so no `railway link` step is required. Each service README gains a short CI/CD note documenting the workflow.

Requires four repo secrets (Railway project tokens): `RAILWAY_TOKEN_PAYMENTS_STAGING`, `RAILWAY_TOKEN_PAYMENTS_PRODUCTION`, `RAILWAY_TOKEN_TRIGGERS_STAGING`, `RAILWAY_TOKEN_TRIGGERS_PRODUCTION`; optional vars `RAILWAY_PAYMENTS_SERVICE` / `RAILWAY_TRIGGERS_SERVICE` override the default service names.

*** NOTE - ONLY STAGING TOKENS HAVE BEEN ADDED TO GITHUB - THIS WILL NOT WORK YET FOR PRODUCTION ***


GTC6244:
Thanks — addressed both points in bc9bc1929:

- **Runner:** switched both workflows to `self-hosted` to match the repo's deploy convention (`deploy-static.yml`, `deploy-staging.yml`, `deploy-prod-1-propose.yml`). Like `deploy-static.yml` reaching Cloudflare, these reach Railway over the public internet from the self-hosted runner.
- **CLI pin:** pinned the Railway CLI to exact `5.57.2` for reproducible deploys. Note the original `@^4` was also wrong — the Railway CLI is on v5, so it would never have resolved a 4.x.

PR #677, opened 2026-09-17
keychain: publish @lit-protocol/keychain to npm on merge to main

clawdbot-glitch003:
## Summary
- **New `keychain-publish.yml`**: on push to `main` touching `lit-agent-keychain/**`, builds the SDK and publishes `@lit-protocol/keychain` with provenance when `sdk/package.json` is ahead of the registry, then pushes the `keychain-sdk-v<version>` tag. Unchanged contents or an already-published version are a no-op, so server-only merges publish nothing.
- **PR gate in `keychain-ci.yml`**: `node scripts/publish-sdk.mjs --check` fails a PR whose packed SDK bytes differ from the published version without a bump. `main` requires reviewed PRs with no bypass actors, so CI cannot commit a bump itself; the bump lives in the PR and the merge is the release.
- **`publish-sdk.mjs`**: adds `--check` and `--no-bump`, skips the interactive `npm login` fallback when `CI` is set. `./publish.sh` locally is unchanged.
- README publishing section updated.

## Registry auth (one-time setup, pick one)
1. **Trusted publishing (recommended, no secret to rotate)**: npmjs.com → `@lit-protocol/keychain` → Settings → Trusted Publisher → GitHub Actions: org `LIT-Protocol`, repo `chipotle`, workflow filename `keychain-publish.yml`, environment blank.
2. **Token fallback**: granular access token, Packages and scopes → Read and write → only `@lit-protocol/keychain`, Bypass 2FA enabled, no org permissions. Save as repo secret `NPM_TOKEN`. npm caps granular tokens at 90 days, so this needs rotating.

## Verified locally
- `--check` on current main: registry already has these exact contents at 2.0.4 → exit 0
- `--check` after touching `sdk/README.md`: fails with the bump instruction
- `--no-bump --dry-run` with version bumped to 2.0.5 under `CI=1`: builds and dry-run publishes 2.0.5, no login prompt


clawdbot-glitch003:
Also restores the nested `zod@3.25.76` lockfile entries that #676 dropped (same fix as #667). npm 10 on the runner rejects the lockfile without them, which is why this PR, `chore/bump-2` and `keychain-homepage-slim-docs-move` all failed at `npm ci`. Validated with `npx npm@10 ci --dry-run`.

